PR #7, opened 2018-12-05
Enable and mandate the use of SSL

atzannes:
# Purpose
We want to use SSL to encrypt communications between the server and the workers (its clients). We only care about encrypting the communication, and not about making sure that the server authenticates with the clients. (This allows a man-in-the-middle attack, but we don't care about that at the moment).

# Approach
the `TcpServer()` object constructor now requires an SSL context. We create such an SSL context from a self-signed SSL cert. The `connect` method didn't need to change because, for the time being, we just want encryption and not authentication, therefore the clients generate a permissive SSL context which does not authenticate the server.

# Testing
All tests pass

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,8 @@
 __pycache__
 build/
 nativepython.egg-info
-.python-version
+
+# leading slash means only match at repo-root level
+/.python-version
+/testcert.key
+/testcert.cert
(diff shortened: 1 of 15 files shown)


atzannes:
@braxtonmckee let me know if it looks good and I will squash before merging because the history of this PR is a mess and not a useful one.

PR #8, opened 2018-12-05
Require TcpServer to be instantiated with SSL & allow to auto-gen that cert

atzannes:


object_database/frontends/database_server.py
```diff
@@ -19,13 +19,17 @@
...
 from object_database.persistence import InMemoryPersistence, RedisPersistence
 
+
 def main(argv):
     parser = argparse.ArgumentParser("Run an object_database server")
 
     parser.add_argument("host")
     parser.add_argument("port", type=int)
+    parser.add_argument("--ssl-path", default=None, required=False, help="path to (self-signed) SSL certificate")
     parser.add_argument("--redis_port", type=int, default=None)
```
atzannes: Underscores in CLI flag names is an anti-pattern. The python argparse knows to store the value in `redis_port` for the flag `--redis-port`.

PR #9, opened 2018-12-07
Alex passwd

atzannes:


object_database/database_connection.py
```diff
@@ -490,6 +499,12 @@ def _noViewsOutstanding(self):
         with self._lock:
             return not self._versioned_data._version_number_refcount
 
+    def authenticate(self, token):
+        # TODO: do I need to hold self._lock?
```
atzannes: @braxtonmckee some of these classes that have locks, but it is not clear what these locks protect, so I'm not sure when to hold one. This is one such case.

object_database/frontends/service_entrypoint.py
```diff
@@ -50,19 +49,23 @@ def main(argv):
...
+        manager = ServiceWorker(
+            dbConnectionFactory, parsedArgs.instanceid,
+            parsedArgs.storageRoot, parsedArgs.serviceToken
+        )
 
         manager.runAndWaitForShutdown()
 
         return 0
-    except:
+    except Exception:
```
atzannes: @braxtonmckee a naked `except` catches both `Exception` and `KeyboardInterrupt`, but I figured we only want to be catching Exceptions in most of these cases.

object_database/schema.py
```diff
@@ -12,30 +12,36 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
+import inspect
+
 from object_database.object import DatabaseObject, Index, Indexed
 from types import FunctionType
-from typed_python import *
```
atzannes: trying to avoid the `from X import *` style which is considered dangerous and error-prone

atzannes:
@braxtonmckee ok I made the changes we talked about. Let me know if things look good now

atzannes:
Rebased and merged manually

PR #13, opened 2018-12-14
breaking _type.cc down a bit

atzannes:
@braxtonmckee what do you think of this initial breakdown? It doesn't help with compilation time yet, but if you feel it's a good start we can merge it in.


braxtonmckee:
yeah I think this is a good step. Breaking all of this into smaller files would be smart. the wrapper class can have a bunch of its functions go into a separate .cc file also...

PR #14, opened 2018-12-18
SerializationContext fixups

atzannes:
1. Assert that we are not initializing the context with an empty name
2. Assert that the names are strings
3. Query the user-defined names before the `builtins`. This should not matter because we the builtin names start with a dot (.), but it is still preferable.

### DIFF
--- a/typed_python/SerializationContext.py
+++ b/typed_python/SerializationContext.py
@@ -31,7 +31,9 @@ _builtin_name_to_value[".builtin._ndarray"] = _ndarray
 _builtin_name_to_value[".builtin.dtype"] = numpy.dtype
 _builtin_name_to_value[".ast.Expr.Lambda"] = Expr.Lambda
 _builtin_name_to_value[".ast.Statement.FunctionDef"] = Statement.FunctionDef
+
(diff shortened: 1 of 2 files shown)


typed_python/SerializationContext.py
```diff
@@ -41,24 +43,42 @@ def __init__(self, nameToObject=None):
...
+            return _builtin_value_to_name.get(tid)
 
     def objectFromName(self, name):
-        if name in _builtin_name_to_value:
-            return _builtin_name_to_value[name]
+        ''' Return an object for an input name(string), or None if not found. '''
+        res = self.nameToObject.get(name)
 
-        return self.nameToObject.get(name)
+        if res:
```
braxtonmckee: should be `res is None` or it will filter out objects that have boolean conversin to false

atzannes: yes! Good catch

atzannes:
I changed the test from `if res` to `if res is not None`. Not sure what tests to add if any. The tests in `test_bad_serialization_context` seem to cover things.

PR #15, opened 2018-12-18
Fixes to Serialization

atzannes:
* Additional serializaiton tests
* Support for `set()` and `frozenset()`
* Refactorings

### DIFF
--- a/typed_python/PythonSerializationContext.hpp
+++ b/typed_python/PythonSerializationContext.hpp
@@ -4,6 +4,39 @@
 #include "SerializationContext.hpp"
 #include "native_instance_wrapper.h"
 
+// PySet_CheckExact is missing from the CPython API for some reason
+#ifndef PySet_CheckExact
+  #define PySet_CheckExact(obj)        (Py_TYPE(obj) == &PySet_Type)
+#endif
+
+
+static inline void throwDerivedClassError(std::string type) {
+    throw std::runtime_error(
+        std::string("Classes derived from `" + type + "` cannot be serialized")
+    );
+}
+
+// Wrapping this macro with a function so we can use it in templated code
+inline PyObject* PyList_Get_Item_No_Checks(PyObject* obj, long idx) {
+    return PyList_GET_ITEM(obj, idx);
+}
+
+// Wrapping this macro with a function so we can use it in templated code
+inline void PyList_Set_Item_No_Checks(PyObject* obj, long idx, PyObject* item) {
+    PyList_SET_ITEM(obj, idx, item);
+}
+
+// Wrapping this macro with a function so we can use it in templated code
+inline PyObject* PyTuple_Get_Item_No_Checks(PyObject* obj, long idx) {
+    return PyTuple_GET_ITEM(obj, idx);
+}
+
+// Wrapping this macro with a function so we can use it in templated code
+inline void PyTuple_Set_Item_No_Checks(PyObject* o, long k, PyObject* item) {
+    PyTuple_SET_ITEM(o, k, item);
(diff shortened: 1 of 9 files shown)


typed_python/PythonSerializationContext.hpp
```diff
@@ -153,89 +236,61 @@ class PythonSerializationContext : public SerializationContext {
...
+        if (!set) {
+            return set;
+        }
 
+        PyObject* res = nullptr;
         try {
-            for (long k = 0; k < sz; k++) {
-                PyTuple_SET_ITEM(res, k, deserializePythonObject(b));
-            }
+            res = PyFrozenSet_New(PyObject_GetIter(set));
```
braxtonmckee: don't forget to decref the iterator, and also to check for exceptions during set construction

typed_python/PythonSerializationContext.hpp
```diff
@@ -655,6 +713,105 @@ class PythonSerializationContext : public SerializationContext {
...
+        }
+
+        size_t sz = b.read_uint32();
+
+        res = factory_fn(NULL);
+        b.addCachedPointer(id, incref(res), true);
+
+        try {
+            for (long k = 0; k < sz; k++) {
+                int success = add_item_fn(res, deserializePythonObject(b));
```
braxtonmckee: are we decreffing the result of 'deserializePythonObject' anywhere?

atzannes: Good question! So `add_item_fn` is `PySet_Add`. It's documentation does not specify if it steals the reference to the key being added to the set. Theoretically it should to keep symmetry with `PyList_SetItem` and `PyTuple_SetItem`.

atzannes: Baaaah!! Looking at the code it looks like `PySet_Add` does not steal the reference. It does `Py_INCREF(key)`. Bogus!

typed_python/PythonSerializationContext.hpp
```diff
@@ -655,6 +739,113 @@ class PythonSerializationContext : public SerializationContext {
...
+            throw std::runtime_error(std::string(
+                "Failed to allocate storage into which to deserialize iterable"));
+        }
+        b.addCachedPointer(id, incref(res), true);
+
+        try {
+            for (long k = 0; k < sz; k++) {
+                PyObject* item = deserializePythonObject(b);
+                int success = add_item_fn(res, item);
+                Py_DECREF(item);
```
atzannes: ok, making sure to decref the item after adding it to a set

typed_python/PythonSerializationContext.hpp
```diff
@@ -216,14 +272,38 @@ class PythonSerializationContext : public SerializationContext {
...
+                // currently being deserialized. In that case PySet_Add will fail, so we temporarily
+                // decref it.
+                auto refcount = Py_REFCNT(res);
+                for (int i = 1; i < refcount; i++) Py_DECREF(res);
+
+                int success = PySet_Add(res, item);
+
+                for (int i = 1; i < refcount; i++) Py_INCREF(res);
+
+                Py_DECREF(item);
```
atzannes: `PySet_Add` does not steal the reference, so we must decref it

atzannes:
@braxtonmckee this is ready for another look from you. I added a number of serialization tests that are failing (which I marked with `@unittest.skip`) so that we have an idea of some of the things that pickle does that we're not, but not have to implement them right-away.

atzannes:
@braxtonmckee this is the second PR. I've rebased it on top of #16

PR #16, opened 2018-12-26
Makefile manages local virtual environment

atzannes:
Using `virtualenv` to manage virtual environment. `pipenv` operates inside the manually created virtual environment.

With this we should be able to stop using `test.py` to build the project.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,18 +1,36 @@
-install: testcert.cert testcert.key
-	pipenv install --system --deploy
+# Path to python binary
+PYTHON ?= $(shell which python3)
 
-test: testcert.cert testcert.key
+# Path to virtual environment
+VIRTUAL_ENV ?= '.venv'
+
+$(VIRTUAL_ENV): $(PYTHON)
+	virtualenv $(VIRTUAL_ENV) --python=$(PYTHON)
+
+.PHONY: install
+install: $(VIRTUAL_ENV)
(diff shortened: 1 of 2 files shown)


atzannes:
@braxtonmckee this is the first PR

Makefile
```diff
@@ -1,18 +1,36 @@
...
+
+.PHONY: install
+install: $(VIRTUAL_ENV)
+	. $(VIRTUAL_ENV)/bin/activate; \
+		pip install pipenv==2018.11.26; \
+		pipenv install --deploy
+
+.PHONY: test
+test: testcert.cert testcert.key install
+	. $(VIRTUAL_ENV)/bin/activate; \
```
braxtonmckee: is there a way to pass command line flags through?

Makefile
```diff
@@ -1,18 +1,36 @@
...
 	./test.py -s
 
+.PHONY: docker-build
```
braxtonmckee: we should probably merge 'build' into this file. no reason to have two entrypoints...

Makefile
atzannes: Yes. It has to be set up explicitly, but the mechanism is the same as the one I used with $(VIRTUAL_ENV) and $(PYTHON)

Makefile
atzannes: Will do

PR #17, opened 2018-12-26
break down compilation

atzannes:
This is on top of the branch `alex-piclex`, which is on top of the branch `alex-venv`.

@braxtonmckee  This is how far I got today breaking down the code into compilation units. No need to merge unless you're planning on working on this code.


typed_python/Instance.hpp
```diff
@@ -0,0 +1,166 @@
+#pragma once
```
braxtonmckee: if you break this into .cc and .h, the .h can fwd declare Type and won't need to include Type.hpp.

atzannes: I'll give that a shot. 

Separately, I tried to break `Type.hpp` to multiple files but got stuck because of the `check` method which requires knowing its child classes. If you think there is a way around that we can look into it when you're back next week

typed_python/Instance.hpp
```diff
@@ -0,0 +1,166 @@
...
+
+        l->refcount = 1;
+        l->type = t;
+
+        mLayout = l;
+    }
+
+    template<class initializer_type>
+    Instance(Type* t, const initializer_type& initFun) : mLayout(nullptr) {
+        t->assertForwardsResolved();
```
atzannes: not sure what to do with this constructor. If I put it in the `cpp` file, `_types.cc` which calls it will not find the correctly instantiated template. If I keep it in the `hpp` file, we cannot just have the forward declaration for `class Type`. We can look at this when you're back.

PR #18, opened 2019-01-02
Members with default values

atzannes:
# Purpose 
To allow us doing things like this:
```
class MyClass(Class):
    x = Member(float, 3.14)
    y = Message(str, "Here be dragons")
```

# Approach
`HeldClass` holds an `Instance` along with each member `Type` and we unwrap `PyObject*` that comes from the python side to an `Instance` object on the C side.

# Questions
- `native_instance_wrapper::tryUnwrapPyObjectToInstance` should probably be `static Instance Instance::create(PyObject* o)` , right?

# Notes
- Renamed `isTypedPythonType`  to `isNativeType` to follow more closely the naming style of the rest of the code.

### DIFF
--- a/typed_python/Type.hpp
+++ b/typed_python/Type.hpp
@@ -8,6 +8,7 @@
 #include <mutex>
 #include <set>
 #include <map>
+#include <tuple>
 #include <utility>
 #include <atomic>
 #include <iostream>
@@ -1958,24 +1959,24 @@ public:
     template<class visitor_type>
     void _visitContainedTypes(const visitor_type& visitor) {
         for (auto& o: m_members) {
-            visitor(o.second);
+            visitor(std::get<1>(o));
         }
     }
 
     template<class visitor_type>
     void _visitReferencedTypes(const visitor_type& visitor) {
         for (auto& o: m_members) {
-            visitor(o.second);
+            visitor(std::get<1>(o));
         }
         for (auto& o: m_memberFunctions) {
-            Type* t = o.second;
+            Type* t = std::get<1>(o);
             visitor(t);
-            assert(t == o.second);
+            assert(t == std::get<1>(o));
         }
         for (auto& o: m_staticFunctions) {
-            Type* t = o.second;
+            Type* t = std::get<1>(o);
             visitor(t);
-            assert(t == o.second);
+            assert(t == std::get<1>(o));
         }
     }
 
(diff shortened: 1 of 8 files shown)


typed_python/native_instance_wrapper.cc
```diff
@@ -2915,11 +2989,17 @@ Type* native_instance_wrapper::unwrapTypeArgToTypePtr(PyObject* typearg) {
             return PythonSubclass::Make(nativeT, pyType);
         } else {
             Type* res = native_instance_wrapper::extractTypeFrom(pyType);
-            return (res) ? res : PythonObjectOfType::Make(pyType);
+            if (res) {
```
atzannes: This is me just being more verbose for clarity

atzannes:
@braxtonmckee I have a working version of Member default value init for some basic types. Take a look and let me know if things look right. If you're happy with the limited types that are currently supported, I'm happy to leave things in this state after I do some minor code refactoring (see question).

typed_python/Type.hpp
```diff
@@ -1146,15 +1146,6 @@ class Bytes : public Type {
 
 class Value : public Type {
 public:
-    Value(Instance instance) :
```
atzannes: This should be `private` right? (Because we want to always use `Value::Make` to create singleton values)

typed_python/native_instance_wrapper.cc
```diff
@@ -2868,57 +2868,34 @@ Instance native_instance_wrapper::tryUnwrapPyObjectToInstance(PyObject* inst) {
...
-        return Value::MakeString(
-            kind == PyUnicode_1BYTE_KIND ? 1 :
-            kind == PyUnicode_2BYTE_KIND ? 2 :
-                                            4,
-            PyUnicode_GET_LENGTH(typearg),
-            kind == PyUnicode_1BYTE_KIND ? (char*)PyUnicode_1BYTE_DATA(typearg) :
-            kind == PyUnicode_2BYTE_KIND ? (char*)PyUnicode_2BYTE_DATA(typearg) :
-                                           (char*)PyUnicode_4BYTE_DATA(typearg)
-            );
+    Instance inst = unwrapPyObjectToInstance(typearg);
```
atzannes: call `unwrapPyObjectToInstance` because it implements most of what we need here.

typed_python/native_instance_wrapper.h
```diff
@@ -183,7 +183,7 @@ struct native_instance_wrapper {
 
     static PyObject* categoryToPyString(Type::TypeCategory cat);
 
-    static Instance tryUnwrapPyObjectToInstance(PyObject* inst);
+    static Instance unwrapPyObjectToInstance(PyObject* inst);
```
atzannes: drop the `try` from the name because we are setting the error if unable to unwrap (so it's not a graceful try method)

atzannes:
@braxtonmckee I did some refactoring. Ready to merge if you are ok with the changes

PR #21, opened 2019-01-08
Add login support to our web interface

atzannes:
# Purpose
Login with LDAP 

Note: use the `ldap_ntlm_domain` option to login with the `sAMAccountName`.

### DIFF
--- a/object_database/web/ActiveWebService.py
+++ b/object_database/web/ActiveWebService.py
@@ -27,21 +27,28 @@ import os
 import json
 import gevent.socket
 
-from gevent import pywsgi
+from object_database.util import genToken, checkLogLevelValidity
+from object_database import ServiceBase, service_schema, Schema, Indexed, Index, DatabaseObject
+from object_database.web.AuthPlugin import AuthPluginBase, PermissiveAuthPlugin, LdapAuthPlugin
+from object_database.web.cells import *
+from typed_python import TupleOf
+
+from gevent import pywsgi, sleep
+from gevent.greenlet import Greenlet
 from geventwebsocket.handler import WebSocketHandler
 
-from object_database.web.cells import *
-
-from gevent.greenlet import Greenlet
-from gevent import sleep
-
-from flask import Flask, send_from_directory, redirect, url_for, request
+from flask import Flask, send_from_directory, redirect, url_for, request, render_template, flash
+from flask_wtf import FlaskForm
 from flask_sockets import Sockets
 from flask_cors import CORS
+from flask_login import LoginManager, current_user, login_user, logout_user, login_required
 
-from object_database import ServiceBase, service_schema, Schema, Indexed, Index, DatabaseObject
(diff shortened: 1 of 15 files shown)


object_database/web/ActiveWebService.py
```diff
@@ -99,12 +226,75 @@ def doWork(self, shouldStop):
...
+        self.app.add_url_rule('/services/<path:path>', endpoint=None, view_func=self.sendPage)
+        self.app.add_url_rule('/login', endpoint=None, view_func=self.login, methods=['GET', 'POST'])
+        self.app.add_url_rule('/logout', endpoint=None, view_func=self.logout)
+        self.sockets.add_url_rule('/socket/<path:path>', None, self.mainSocket)
+
+    def login_user(self, username, remember=False):
+        with self.db.transaction():
+            users = User.lookupAll(username=username)
+
+            if len(users) == 0:
```
braxtonmckee: this lookup-or-create pattern ought to have a 'revisionConflictRetry' around it.

object_database/web/ActiveWebService.py
```diff
@@ -27,34 +27,106 @@
...
+
+
+@active_webservice_schema.define
+class User:
+    username = Indexed(str)
+
+    def __init__(self, username):
+        self.username = username
+
+    def is_authenticated(self):
```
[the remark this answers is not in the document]
atzannes: that was my bad understanding. It actually should return `True` or `False` based on whether the user is authenticated, which I am not making use of.

object_database/web/ActiveWebService.py
atzannes: :+1:

object_database/web/ActiveWebService.py
```diff
@@ -27,34 +27,130 @@
...
+        else:
+            return UserWrapper(user.username, user.login_expiration)
+
+    def __init__(self, username, login_expiration):
+        self.username = username
+        self.login_expiration = login_expiration
+
+    @property
+    def is_authenticated(self):
+        return True if time.time() < self.login_expiration else False
```
braxtonmckee: you might want to have the 'db' here instead of the login_expiration, and check the value directly from a view rather than stashing it in the object. Otherwise, this object will become stale. If flask is not stashing this anywhere (that is, if it's recreating this object every time we use it) then it won't be a problem. But as it stands, anyone who has this object could get an out-of-date view of the ttl if they just hold onto the object.

atzannes: I though about that but the model is that the `flask-login` plugin assumes that the User objects live in some sort of DB (typically a relational DB) and that's why it requires you to register a callback with it for `load_user` that it calls each time it needs to access a User field, for example `is_authenticated` (I checked that by adding print statements to the `load_user` call back and the `is_authenticated` property). So, I thought that this dumb UserWrapper design is better because it is simpler.

PR #25, opened 2019-01-16
Use PERFORMANCE_FACTOR throughout `database_test.py` tests

atzannes:
Things take longer on Travis and the `PERFORMANCE_FACTOR` had been
implemented for the tests that were failing the most, but another one
failed today, so I applied the same concept to the remaining tests in that
file, which seems to be making Travis happy


atzannes:
@braxtonmckee FYI I did this

PR #29, opened 2019-01-22
Some small fixups

atzannes:


object_database/frontends/service_config.py
```diff
@@ -50,7 +50,7 @@ def _main(argv):
 
     parser.add_argument("--hostname", default=os.getenv("ODB_HOST", "localhost"), required=False)
     parser.add_argument("--port", type=int, default=int(os.getenv("ODB_PORT", 8000)), required=False)
-    parser.add_argument("--auth", type=str, default=os.getenv("ODB_AUTH_TOKEN"), required=False, help="Auth token to use to connect.")
+    parser.add_argument("--auth", type=str, default=os.getenv("ODB_AUTH_TOKEN", ""), required=False, help="Auth token to use to connect.")
```
atzannes: Make the default the empty string rather than `None`. Otherwise downstream code fails saying it can't convert `None`  to a string, instead of saying something like "wrong token" when it tries to connect.

PR #34, opened 2019-02-04
Codebase fixups

atzannes:
- `rootlevelPathFromModule` needed a fixup to work correctly when the input is a root module (such as `typed_python`)
- moving an `import` inside the function that needs it in order to avoid cyclic import problems (e.g., we could not `import typed_python.Codebase` without this change).


atzannes:
This is ready for review @braxtonmckee

PR #39, opened 2019-02-15
ran pipenv lock to get fresh packages

atzannes:
# Purpose
There was a security vulnerability in the version of numpy we were using.

# Approach
- `pipenv lock` to get updated dependencies
- `pipenv check` on Travis to get failed builds when new security vulnerabilities are discovered

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,11 +1,11 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "1399d818904908ad87d0fa165abddf9248b09dcef860a31255ba6e3aefaffe06"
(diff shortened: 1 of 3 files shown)


Pipfile.lock
```diff
@@ -1,11 +1,11 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "1399d818904908ad87d0fa165abddf9248b09dcef860a31255ba6e3aefaffe06"
+            "sha256": "fbb0f5c3501780db69c92a1a1a7fa01b1b1059eebd5ee11d09a414d9357eb496"
         },
         "pipfile-spec": 6,
         "requires": {
-            "python_version": "3.6.7"
+            "python_version": "3.6"
```
atzannes: `pipenv check` didn't like having the `.7` in the python version. It prevented it from doing the check.

PR #44, opened 2019-02-25
Lint: Some semantic fixes (& some more cosmetic stuff)

atzannes:
## Overview
After these changes, flake8 complains about the following:
```
570   E501 line too long (110 > 99 characters)
3     E721 do not compare types, use 'isinstance()'
39    E741 ambiguous variable name 'l'
406   F401 'nativepython.runtime.Entrypoint' imported but unused
30    F403 'from typed_python import *' used; unable to detect undefined names
421   F405 'Alternative' may be undefined, or defined from star imports: typed_python
1     F632 use ==/!= to compare str, bytes, and int literals
29    F811 redefinition of unused '__init__' from line 63
8     F821 undefined name 'out'
2     F841 local variable 'native_output_type' is assigned to but never used
16    W503 line break before binary operator
65    W504 line break after binary operator
1590
```

## TODO [Done]
 Fixed most "F841: local variable 'xyz' is assigned to but never used" errors, but  a couple of instances remain:
```
./nativepython/python_to_native_converter.py:79:13: F841 local variable 'native_output_type' is assigned to but never used
            native_output_type = output_type.getNativeLayoutType()
            ^
./nativepython/python_to_native_converter.py:161:13: F841 local variable 'body' is assigned to but never used
            body = [python_ast.Statement.Return(
            ^
```

### DIFF
--- a/nativepython/expression_conversion_context.py
+++ b/nativepython/expression_conversion_context.py
@@ -23,10 +23,10 @@ from nativepython.type_wrappers.none_wrapper import NoneWrapper
 from nativepython.python_object_representation import pythonObjectRepresentation, typedPythonTypeToTypeWrapper
 from nativepython.typed_expression import TypedExpression
 from nativepython.conversion_exception import ConversionException
+from typed_python import *
+
 
 NoneExprType = NoneWrapper()
-
-from typed_python import *
 
 typeWrapper = lambda t: nativepython.python_object_representation.typedPythonTypeToTypeWrapper(t)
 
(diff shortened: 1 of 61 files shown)


atzannes:
@braxton, take a look at the PR TODO: I didn't want to remove this dead code in case the error was that we forgot to use those unused local variables.

braxtonmckee:
In the case of the TODOs, the first one you can see the calculation is being replicated lower down. We should just use the variable. In the second one, that looks like a bug we're just not hitting in test cases. we should be using 'body', not pyast.body.

PR #50, opened 2019-03-11
Replace from ... import * by importing the objects and methods which are actually used.

atzannes:
Wait for the Lint test to pass, then merge

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,6 @@ install:
 
 script:
   - make testcert.cert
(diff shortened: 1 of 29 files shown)


atzannes:
@braxtonmckee  check out the "Fixups" commit. I fixed some "undefined" errors in a way that seemed reasonable, and tests are passing, but a 2nd pair of eyes would be helpful because I suspect these lines are not being tested.

atzannes:
(got a verbal :+1: , so merging)

PR #53, opened 2019-03-13
Enforcing linting for the lint-codes that I have fixed

atzannes:
# Purpose
I saw some lint warnings categories that I had fixed creeping back in. It's probably a good idea to keep our code looking sharp and to be individually responsible for keeping it so.

# Approach
- Added TravisCI tests for the lint-codes that I have fixed
- Added `make` rules  for linting that do not require you to be using the virual environment. Just do `make lint` from the root directory to check for errors, or `make vlint` if building using the virualenvironment.


atzannes:
@braxtonmckee the lint police strikes back (that's me!) Ready to merge with your blessings :)

dkrasner:
@atzannes this looks good; i stick with general py3 standards when writing. As I work on pieces of the codebase and see style errors should I go ahead and change, or do you have a plan to clean up/lint the entire code base and I should just leave as is.

atzannes:
@dkrasner I do plan to lint the entire code-base, with the exception of some codes listed in `tox.ini`, which we may revisit at some point. The `make lint-all` rule (on this branch) shows the lint errors I have yet to tackle. So, regarding whether to lint-as-you-go, I would say if it doesn't bother you, hold off for now so that it doesn't become a distraction and so that the PR diffs are not "polluted" by lint fixes.

PR #59, opened 2019-03-15
Compile str.find()

rubidge:
## Motivation and Context
* New feature: Compile str.find() to C++

## Approach
* Implemented C++ functions on type String, one for each possible number of parameters, e.g. str.find(s, sub), str.find(s, sub, start), and str.find(s, sub, start, end)
* Defined corresponding runtime functions to call from python, and when compiling, create the appropriate call in String wrapper, according to the number of arguments provided.
## How Has This Been Tested?
* Test on empty, small, and large base strings and substrings.
* Vary start and end over a wide range of values.
* Test too few and too many arguments. Generate TypeError in these cases.

###Note
* Provides same functionality, but not necessarily performant, at least in preliminary tests.

### DIFF
--- a/nativepython/tests/string_compilation_test.py
+++ b/nativepython/tests/string_compilation_test.py
@@ -27,9 +27,9 @@ someStrings = [
     "a",
     "as\x00df",
     "\u00F1",
-    "\u0F01",
-    "\u0F01",
-    "\u1002"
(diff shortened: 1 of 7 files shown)


rubidge:
Will correct flake8 errors

PR #61, opened 2019-03-19
Fix and Enforce F821: undefined name 'xyz'

atzannes:
## Motivation and Context
Undefined variables are usually bugs: we should not have them or in the rare cases when they're not bugs (e.g., in test code), we should flag them as intentional.

## Approach
- Enforce checking for F821 errors
- Fix cases where we had undefined variables
- Mark cases where undefined variables were intentional (for testing)

## How Has This Been Tested?
- [x] Waiting for TravisCI tests to pass
- [ ] Waiting for Braxton to give :+1: on the code changes

## Types of changes

## Checklist:


atzannes:
@braxtonmckee I thought I had already taken care of undefined variables (F821), but apparently I had only taken care F811 (redefinition of unused variables). Most of the changes were `# noqa: F821` in test files, but there were a few in nativepython-wrappers which seemed like bugs. I fixed them in a way that seemed reasonable, but I'm not familiar with that code, so please take a look.

atzannes:
@braxtonmckee can you take a look at the code changes when you get some time?

PR #65, opened 2019-03-21
Finalize Lint enforcement

atzannes:
## Motivation and Context
Strengthen the way in which we use `flake8` in order to maintain a unified code-style. 

Until now, we had an opt-in approach to enforcing code-style: we had selected a list of codes (e.g., F402, etc) that were being checked when running `flake8`. This PR expands the number of those codes being checked. In fact, with this PR we are switching from opt-in to opt-out, where ALL `flake8` warnings are produced, except those for wich we have opted-out in the `tox.ini` file.

## Approach
- I went through the list of error codes we had not enabled, confirmed that they made sense, enabled them and fixed any issues
- Replaced the file-level `noqa` directives with a `per-file-ignore` rule that excludes the single code rather than all codes for those files (`__init__.py` files are allowed to import names without using them in order to define a module API).
- Replaced all bare `noqa` with qualified `noqa` (e.g., ` # noqa: F821`).

## How Has This Been Tested?
TravisCI tests pass

## Checklist:


atzannes:
FYI @dkrasner @darth-cheney the lint police strikes again!

PR #66, opened 2019-03-22
HTML Generation

darth-cheney:
The PR primarily deals with separating HTML string generation from any other object that spits out HTML.

## Motivation and Context

In order to keep classes like Cells and friends as simple as possible, and to prevent easy errors on the part of programmers, it's a good idea to avoid writing HTML by hand.
  

With the classes introduced in this PR, we can now create elements in Python without having to worry about how the strings look or how to print children, etc.

## Approach

The HTML generation classes and their tests can be found in `object_database/web/html`
  
### HTML String Generation ###
The idea here is fairly simple: any class that subclasses `AbstractHTMLWriter` (for now this is only `HTMLElement` and `HTMLTextContent`) must implement a method called `print_on` whose first argument is some kind of stream.
  
In our case we are using `StringIO` to build strings in a streamlike fashion.
  
When `__str__` is called on any `HTMLElement` instance, it will create a new stream and pass it to its own `print_on`, which handles printing open tags, attributes, children, and close tags. Printing children is pretty simple: it will just call `print_on` and pass the working stream to each of its children in order, and so on recursively down the chain.
  
To test this out, try something like this:
```python
parent = HTMLElement('div')
parent.add_class("container")
child_one = HTMLElement('img', is_self_closing=True)
child_two = HTMLElement('p')
child_content = HTMLTextContent("This is the paragraph content")
child_two.add_child(child_content)
parent.add_children([child_one, child_two])
print(parent)
```
Now this code is somewhat verbose, which is why we added other ways to do this (see chaining and constructors sections below)
  
### Chaining ###
The example in the section above shows how writing this HTML can be fairly explicit. But we've added the convenience of chaining most public methods (ie, returning `self`) so that we can do things like:
```python
element = HTMLElement('div').with_children(
    HTMLElement('img', is_self_closing=True).add_class("primary-img"),
    HTMLElement('p').add_class("main-typography").add_child(
        HTMLTextContent("This is the paragraph content")
    )
)
print(element)
```
This allows for some better composition.
  
### Special Constructors ###
Some HTML5 elements are "self-closing," meaning they don't have any children nor to they use a totally separate closing tag. For example, `<div>Something</div>` is not self closing, while `` **is** self closing.
  
When using the plain class constructor, one would need to know ahead of time whether or not the tag is self closing or not the generate valid HTML5. So we decided to dyamically add class methods to `HTMLElement` that "do the right thing" for you:
  
````python
div = HTMLElement.div().add_class("main")
img = HTMLElement.img().add_class("big")
div.add_child(img)
print(div)
# <div class="main">
#     <img class="big" />
# </div>
````
  
### Cells Proof of Concept ###
We have successfully used these classes as drop-in replacements for the HTML that is generated in
three `Cells` classes: `Card`, `CardTitle`, and `CollapsiblePanel`. So far these appear to load and work fine when running the database test server.
  
## How Has This Been Tested?

We have created a full set of tests in `object_database/web/html/html_gen_tests.py` and all are currently passing.
  
All current Cells tests are also passing, as is the regular full test suite.
  
For testing the output, we run `object_database/frontends/object_database_webtest.py` and then load all relevant services in the browser, checking the console as we go. So far any errors that do appear in the browser console are identical to those on the current (as of this writing) `dev` branch and are non-fatal.
  
We have also run flake8 on the changed files and everything seems to be ok linting wise.

  
## Types of changes

## Checklist:

### DIFF
--- a/object_database/web/cells.py
+++ b/object_database/web/cells.py
@@ -893,7 +893,7 @@ class Card(Cell):
         body = HTMLElement.div().add_class(
             "card-body").add_class(
                 other).add_child(
-                    HTMLTextContent("____contents__"))
+                    HTMLTextContent(" ____contents__"))
         card = HTMLElement.div().add_class("card")
 
         if self.header is not None:
@@ -907,6 +907,8 @@ class Card(Cell):
         card.add_child(body)
         card.attributes["style"] = self._divStyle()
 
+        self.cells._logger.info(str(card))
+        
         self.contents = str(card)
 
     def sortsAs(self):
--- a/object_database/web/cells.py
+++ b/object_database/web/cells.py
@@ -15,6 +15,7 @@ from inspect import signature
 from object_database.view import RevisionConflictException
 from object_database.view import current_transaction
 from object_database.util import Timer
+from object_database.web.html.html_gen import HTMLElement, HTMLTextContent
 
 MAX_TIMEOUT = 1.0
 MAX_TRIES = 10
@@ -885,28 +886,28 @@ class Card(Cell):
     def recalculate(self):
         self.children = {"____contents__": Cell.makeCell(self.body)}
 
-        if self.header is not None:
-            self.children['____header__'] = Cell.makeCell(self.header)
-
         other = ""
         if self.padding:
             other += " p-" + str(self.padding)
 
-        self.contents = ("""
-            <div class="card" __style__>
-              __header__
-              <div class="card-body __other__">
-                ____contents__
-              </div>
-            </div>
-            """.replace('__other__', other)
-               .replace('__style__', self._divStyle())
-               .replace("__header__", """
-                    <div class="card-header">
-                        ____header__
-                    </div>
-                """ if self.header is not None else "")
-        )
+        body = HTMLElement.div().add_class(
+            "card-body").add_class(
+                other).add_child(
+                    HTMLTextContent("____contents__"))
+        card = HTMLElement.div().add_class("card")
+
+        if self.header is not None:
+            header = HTMLElement.div().add_class(
+                "card-header").add_child(
+                    HTMLTextContent("____header__")
+            )
+            self.children['____header__'] = Cell.makeCell(self.header)
+            card.add_child(header)
+
+        card.add_child(body)
+        card.attributes["style"] = self._divStyle()
+
+        self.contents = str(card)
 
     def sortsAs(self):
         return self.contents.sortsAs()
--- a/object_database/web/cells.py
+++ b/object_database/web/cells.py
@@ -908,7 +908,7 @@ class Card(Cell):
         card.attributes["style"] = self._divStyle()
 
         self.cells._logger.info(str(card))
-        
+
         self.contents = str(card)
 
     def sortsAs(self):
@@ -920,11 +920,9 @@ class CardTitle(Cell):
         super().__init__()
 
         self.children = {"____contents__": Cell.makeCell(inner)}
-        self.contents = """
-        <div class='card-title'>
-          ____contents__
-        </div>
-        """
+        self.contents = HTMLElement.div().add_child(
+            HTMLTextContent("____contents__")
+        ).add_class("card-title")
 
     def sortsAs(self):
         return self.inner.sortsAs()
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: object_database/web/cells.py [hunk 1]
```diff
@@ -917,11 +920,9 @@ def __init__(self, inner):
         super().__init__()
 
         self.children = {"____contents__": Cell.makeCell(inner)}
-        self.contents = """
-        <div class='card-title'>
-          ____contents__
-        </div>
-        """
+        self.contents = HTMLElement.div().add_child(
+            HTMLTextContent("____contents__")
```
braxtonmckee: so one of the advantages of your approach is that instead of having this ridiculous ____X__ pattern, we could have an additional html element called 'CellsChild' that is supposed to get swapped out for an appropriate cells child later once its properly built. This could really clean up the code because the references to the child items would be much more obvious.

Probably not for this PR, though.

darth-cheney: Yeah I think there are a few different ways we could merge the functionality together a bit more, whether it's as you've suggested or if every cell has it's own `self.element` that is the main wrapper element for the cell, or if the `Cell` class itself requires a `print_on` (or `printOn`) in similar fashion for any string generation functions. We should discuss it for sure.

dkrasner: We could also have the init  or some other chosen method return `self` and have more dom-js style patterns such as:
```python
HTMElement.div(
      some stuff).p(
           some more stuff)....
```

this would be an anti-pattern of sorts but potentially worth considering given that we are trying to find some balance between the python world and the dom-html-js world. 

We can also add unique ids, and some "DOM/Node" class, that all the HTMLElements inherit, that you can do placeholder by-id replacements. For example:

```python
element = HTMLElement('div', id=1).with_children(
    HTMLElement('img', is_self_closing=True, id=2).add_class("primary-img"),
    HTMLElement('p', id=3).add_class("main-typography").add_child(
        HTMLTextContent("This is a placeholder", id=4)
    )
)

element.replaceByID(id=4, html=HtmlTextContent("We just changed this element"))
```

which would be equivalent to 
```python
element = HTMLElement('div', id=1).with_children(
    HTMLElement('img', is_self_closing=True, id=2).add_class("primary-img"),
    HTMLElement('p', id=3).add_class("main-typography").add_child(
        HTMLTextContent("We just changed this element", id=4)
    )
)
```

these ids could be user defined (semantic or int), auto-generated as default etc 

so a few things to discuss.

object_database/web/cells.py
```diff
@@ -885,28 +906,30 @@ def __init__(self, body, header=None, padding=None):
...
-               .replace("__header__", """
-                    <div class="card-header">
-                        ____header__
-                    </div>
-                """ if self.header is not None else "")
-        )
+        body = HTMLElement.div().add_class(
+            "card-body").add_class(
+                other).add_child(
+                    HTMLTextContent(" ____contents__"))
```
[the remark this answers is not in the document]
braxtonmckee: I also find this easier to read

object_database/web/cells.py
braxtonmckee: So I was merely thinkiing that right now we write '____contents__' inline in the html, and separately have a dictionary with {'____contents__': some_cell} and this is how we register the placement in the DOM of our child cells - it's more explicit and wouldn't require direct text substitution.

object_database/web/cells.py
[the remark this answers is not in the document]
darth-cheney: Yeah. In fact, the format you've proposed @atzannes is what would normally be used in JS. I just haven't encountered "chainable" patterns in Python too often and wasn't sure how to format them properly and have linters pass. I believe you would need a `\` at the end of each line in that construction though, right? Either way the "separate lines" approach is certainly easier to read.

darth-cheney:
Shall we merge and then deal with other formatting issues in later PRs?

braxtonmckee:
The enclosing parens mean you don’t need escape slashes

braxtonmckee:
Yes let’s merge and when you do the rest of the cells we can try to get a consistent style

object_database/web/cells.py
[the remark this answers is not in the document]
atzannes: If I'm not mistaken, we do not need the `\` at the end of the line because we are within a parenthesized expression. Both the linter and python should be happy with that syntax

PR #67, opened 2019-03-25
* Compile str.upper() and str.is...() functions

rubidge:
## Motivation and Context
    These are all the str.is...() functions, except for str.isidentifier().
   
## Approach
* The str.is...() functions require a Unicode property table, which is generated by unicodeprops.py
      * Should this script be invoked in Makefile? or in setup.py?  It's currently invoked in Makefile.

## How Has This Been Tested?
* Created test cases for each new function.  Important cases are null strings, \x00, all codepoint sizes, titlecase codepoints, combinations of properties within string.

## Types of changes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,7 @@ O_FILES = $(BUILD_PATH)/_types.o \
           $(BUILD_PATH)/Type.o \
           $(BUILD_PATH)/PythonSerializationContext.o
 
+UNICODEPROPS = $(SRC_PATH)/UnicodeProps.hpp
 
 ##########################################################################
 #  MAIN RULES
@@ -31,6 +32,7 @@ O_FILES = $(BUILD_PATH)/_types.o \
 install: $(VIRTUAL_ENV)
 	. $(VIRTUAL_ENV)/bin/activate; \
 		pip install pipenv==2018.11.26; \
+		$(PYTHON) $(SRC_PATH)/unicodeprops.py > $(UNICODEPROPS); \
 		pipenv install --dev --deploy
 
 .PHONY: test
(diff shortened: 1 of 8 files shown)


rubidge:
Didn't invoke unicodeprops.py, so unicode properties header file not created, so c++ compilation failed.

nativepython/type_wrappers/string_wrapper.py
```diff
@@ -165,6 +167,116 @@ def convert_method_call(self, context, instance, methodname, args, kwargs):
                         ).cast(self.layoutType)
                     )
                 )
+        elif methodname == "upper":
```
braxtonmckee: might make sense to have a dictionary from method to the runtime function if the pattern is consistent

Makefile
```diff
@@ -31,6 +32,7 @@ O_FILES = $(BUILD_PATH)/_types.o \
 install: $(VIRTUAL_ENV)
 	. $(VIRTUAL_ENV)/bin/activate; \
 		pip install pipenv==2018.11.26; \
+		$(PYTHON) $(SRC_PATH)/unicodeprops.py > $(UNICODEPROPS); \
```
braxtonmckee: Generating UnicodeProps.hpp will either have to be part of setup.py, or done as a step where we check the file into source control (and then we just need a 'make' target to update it periodically).

rubidge: Thanks, will do.

nativepython/type_wrappers/string_wrapper.py
rubidge: Yes, I agree.  Will do.

PR #68, opened 2019-03-25
Compile str.upper() and str.is...() 

rubidge:
## Motivation and Context
    * These are all the str.is...() methods, except for str.isidentifier().

## Approach
* The str.is...() functions require a Unicode property table UnicodeProps.hpp
    * Generate UnicodeProps.hpp when needed, with "make unicodeprops".
* Use dictionary of functions to reduce duplicate code, when implemented methods have the same signature.

## How Has This Been Tested?
* Created test cases for each new function. Important cases are null strings, \x00, all codepoint sizes, titlecase codepoints, combinations of properties within string.

## Types of changes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,7 @@ O_FILES = $(BUILD_PATH)/_types.o \
           $(BUILD_PATH)/Type.o \
           $(BUILD_PATH)/PythonSerializationContext.o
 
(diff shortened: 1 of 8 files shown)


rubidge:
I really do observe the performance difference between dev and this branch.  Not just an accidental failure.  Investigating further.

PR #70, opened 2019-03-26
Compile str.upper() and str.is...()

rubidge:
## Motivation and Context
    * These are all the str.is...() methods, except for str.isidentifier().

## Approach
* The str.is...() functions require a Unicode property table UnicodeProps.hpp
    * Generate UnicodeProps.hpp when needed, with "make unicodeprops".
* Use dictionary of functions to reduce duplicate code, when implemented methods have the same signature.

## How Has This Been Tested?
* Created test cases for each new function. Important cases are null strings, \x00, all codepoint sizes, titlecase codepoints, combinations of properties within string.

## Types of changes


braxtonmckee:
For brevity in the codebase, I wonder if you could compress unicode.hpp using runlength encoding. e.g.

```

uint16_t* initialize_uprops() {
  uint16_t uprops_value = { 0x0, 0x40, 0x0, ...};
  uint16_t uprops_count = { 9, 5, ... };
  uint16_t uprops_size = 5000; //or however many things are here
  
  uint16_t* result = new uint16_t[0x110000];

  long cur_write_slot = 0;
  for (long k = 0; k < uprops_size; k++) {
    for (long j = 0; j < uprops_count[k]; j++) {
      result[cur_write_slot++] = uprops_value[k];
  }

  return result;
}
uint16_t* uprops = initialize_uprops();

```
everything would be a little smaller that way, including the source, since there are large blocks of 0x0 in there etc. Otherwise, looks good, please merge.

rubidge:
I forgot to add the file unicodeprops.py to this branch (it was in jeff-str-is but not jeff-str-is2).  So I'll do the runlength encoding, add unicodeprops.py, and resubmit.

rubidge:
I made a new branch that can rebase without conflict hopefully.

PR #72, opened 2019-03-27
Make cells into a (sub) module.

atzannes:
## Motivation and Context
The goal is to start breaking the huge `cells.py` file into smaller ones, while still importing its functionality in the same easy way (from `object_database.web.cells`). 
This PR is just setting the stage: it does not yet perform any of the breaking of `cells.py` into files.

## Approach
- Move everything into the `cells` subdirectory.
- Import all classes and global methods of cells in the sub-module's `__init__.py` file

## How Has This Been Tested?
This change didn't break the TravisCI tests (hopefully, TBD).


atzannes:
@dkrasner @darth-cheney  what do you guys think of this change? I didn't want to start breaking `cells.py` into multiple files because I know you are working on this surface.

If you think it's a good idea and it's easy for you to rebase on top of this commit, we can merge it.

atzannes:
@braxtonmckee  FYI ^

darth-cheney:
@atzannes This looks good to me. Shouldn't interfere with what I'm in the middle of, which is swapping out the rest of the HTML in Cells with the generator classes. What you have started here def needs to happen too. I think once I submit the subsequent PR we can start breaking things up into smaller modules.

PR #74, opened 2019-03-28
Complete Cells HTML Integration

darth-cheney:
This PR completes the process of an initial integration of HTML generation in Cells.

## Motivation and Context

This completes a process begun with PR #66 

## Approach

For now, we simply swap any HTML declared in string literals with
the HTML Generation python equivalent code.
  
There are also some minor changes to the html generation library. For example, there is a new `set_attribute` chainable method that allows you to set any HTML attribute. This is useful for attributes like `role=` and even the current `onclick=` inline callbacks.
  
We imagine that even what is present in this PR will be subject to change as we break out Cells modules and begin to refactor.

## How Has This Been Tested?

Ran all relevant test suites. I also booted up the test object webserver environment and attempted to test what I could manually. I might have missed something. **Please try this out in a live browser and let me know if you find any errors.** Some of the HTML "swapping" was non-trivial and could have introduced small bugs.

## Types of changes

## Checklist:

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -8,6 +8,7 @@ nose = "*"
 flake8 = "*"
 autopep8 = "*"
 beautifulsoup4 = "*"
+py_w3c = "*"
 
 [packages]
 redis = "*"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "ab74724e765c49e5e02a2727d24dec76cfc3d8d9b09dfa2f0a64397c5cc6706b"
+            "sha256": "95d4a0f0c139d62debf3f7976a7a7a4d8f69009cbedf1dadcc39593787e895f9"
         },
         "pipfile-spec": 6,
         "requires": {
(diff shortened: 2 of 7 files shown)


braxtonmckee:
OK, I'll try running our app on top of this and see how it does.

darth-cheney:
Not sure why CI is failing here -- seems to have something to do with `ServiceManager` even though there were no changes made to that code here.

braxtonmckee:
Try rebasing. I had a linting error that made its way into dev. I pushed a
fix but maybe you didn't rebase on top of it yet.

braxtonmckee:
or no, actually, looks like something is crashing on dev. hold on.

darth-cheney:
@braxtonmckee Ok everything seems to be passing now. I think @dkrasner is probably going to branch from here and do some Cells work

braxtonmckee:
Cool.

Maybe we can write a test that stands up each cell and verifies that the
html it generates is valid and that it doesn't throw an exception?

I still have to try exercising the whole system with this...

darth-cheney:
@braxtonmckee I can definitely write some tests for these. Do you prefer general tests of HTML validity, or instead tests of specific attributes / children being present on / in constiutent HTML elements? I feel like the latter will be changing a lot in the coming weeks...

braxtonmckee:
I think mostly just want to check that the html is valid. Simply exercising
that we didn't throw an exception while implementing the cell would be good.

One pattern that's worked well for me in the past is to have one code path
that generates all the different interesting objects of type 'T' and then
writing tests that work against arbitrary 'T' that express invariants such
as 'no cells should render with exceptions, all the html is valid, etc'.
This can get good coverage without having a trillion little details
embedded in the test that you have to change later.

darth-cheney:
Yeah might be difficult since cells have all sorts of different init requirements but I'll get something up and running that's sufficient

darth-cheney:
@braxtonmckee We should be good to go here. We've added html generation tests and they seem to be passing. The Branch CI build is failing here for some unrelated reason due to the `ServiceManager` test

atzannes:
@darth-cheney, it looks like you've hit one of our flaky tests. I'm not sure of a good way to make them less annoying on TravisCI. Usually, re-running the tests (occasionally more than once) results in a green build.

Separately, it looks like there are some merge conflicts that need to be resolved before this PR can be merged.

dkrasner:
@braxtonmckee local (CLI) rebase and merge works fine, no conflicts and commit history looks good. Should we push to `dev`?

braxtonmckee:
let me try running the app using this branch and verify all is well.

dkrasner:
ok just let us know and we'll push when you are ready..

braxtonmckee:
I squashed the chain, found a couple of minor bugs, and am going to merge once CI completes. 

It was almost impossible to rebase in its broken-out form, but when I squashed it down to the commit you were based on and rebased that it was fine. Everything looks good. I added a test that `__identity__` doesn't show up in any of the generated html/javascript.

PR #80, opened 2019-04-05
ServiceManager can stop a service and wait for it

atzannes:
plus a couple of other small changes:
- renamed isNotRunning to isNotActive because it is much closer to
the negation of the existing isActive method.
- implemented isRunning and used it in ServiceManager.waitRunning


atzannes:
@braxtonmckee the `isRunning` predicate replaces the `instance.state == "Running"` check with something more restrictive. Tests are happy, but can you check that the extra constraints make sense?

PR #81, opened 2019-04-05
Initial JS Cleanup

darth-cheney:
This PR represents a very early cleanup of the JS on the frontend. It has accomplished two goals:
1. Allowed us to get a better understanding of the frontend portion of things by refactoring the code;
2. Allowed us to begin thinking about how a websocket standard protocol should look, how it might interact will `Cells`, and what JS would be required to handle this interaction as flexibly as possible.
  
Note that we are still in the process of step #2, but want to get these changes in.

## Motivation and Context

The Javascript on the front end was improvised in the beginning. We needed to break out the functionality in order to be able to understand future design needs and to make the code a bit more maintainable.

## Approach

We have removed most of the inline Javascript from `page.html` and decomposed functionality out into
two constituent classes: `CellSocket` and `CellHandler`.
  
### `CellSocket` ###
This object gets instantiated in `page.html` as `cellSocket` and therefore is made explicit on the global `window`.
  
`CellSocket` is a kind of wrapper for the underlying `Websocket` object. Because we are using the frame buffering to get around a bug in Chrome, we put this functionality in its own class. We also add convenience methods that allow binding of callbacks for specific normal socket events like `onclose` and `onerror.` However, `CellSocket` also handles the initial parsing of the two current message types: `postscripts` (which is just a textual command) and "normal" messages that are deserialized from JSON and have actual data about DOM replacements.
  
`CellSocket` owns two optional handler callbacks for these two instances. When an incoming message is a `postscripts` type, it will attempt to call its `postscriptsHandler` if it has one. Similarly, if it is of a "normal" type, it will attempt to call its `messageHandler` if it has one.
  
As you might guess, these handlers currently live in `CellHandler`.
  
### `CellHandler` ###
This class deals with handling the two current primary message types. It also holds its own array of JS scripts from the backend (`postscripts`) and a dictionary of cell objects that will be used for tracking replacements and cell updates to the DOM as they come in over the wire through "normal" messages.

### `page.html` ###
The code in the main HTML page is now much shorter. Instantiating the whole socket/handler
combination looks like this:
```javascript
 const cellSocket = new CellSocket();
 const cellHandler = new CellHandler();
 cellSocket.onPostscripts(cellHandler.handlePostscript);
 cellSocket.onMessage(cellHandler.handleMessage);
 cellSocket.onError(err => {
     console.error("SOCKET ERROR: ", err);
 });
 document.addEventListener('DOMContentLoaded', () => {
     cellSocket.connect();
 });
```

## How Has This Been Tested?
We have tested by using the demo frontend in `object_database/frontends` and making sure that interaction works as normal and that no errors are thrown.
  
We have not written any specific unit or integration tests yet, since we expect much of this API to be heavily modified and expanded in the coming couple of weeks as we flesh out the websocket standard.

## Note ##
We are still experimenting with different socket standard ideas, so this code will change a lot in the future. This is just supposed to be a temporary and minor cleanup.

## Types of changes

## Checklist:

### DIFF
--- a/object_database/web/cells/cells.py
+++ b/object_database/web/cells/cells.py
@@ -1333,7 +1333,7 @@ class _NavTab(Cell):
 
     def recalculate(self):
         inlineScript = """
-        websocket.send(JSON.stringify({'event': 'click', 'ix': __ix__, 'target_cell': '__identity__'}))
+        cellSocket.sendString(JSON.stringify({'event': 'click', 'ix': __ix__, 'target_cell': '__identity__'}))
         """.replace('__identity__', self.target).replace('__ix__', str(self.index))
         navLinkClasses = ['nav-link']
         if self.index == self.slot.get():
@@ -1444,7 +1444,7 @@ class Dropdown(Cell):
             self.children["____child_%s__" % i] = Cell.makeCell(header)
             if not isinstance(onDropdown, str):
                 inlineScript = """
-                websocket.send(JSON.stringify({'event':'menu', 'ix': __ix__, 'target_cell': '__identity__'}))
+                cellSocket.sendString(JSON.stringify({'event':'menu', 'ix': __ix__, 'target_cell': '__identity__'}))
                 """.replace('__ix__', str(i)).replace('__identity__', self.identity)
             else:
                 inlineScript = quoteForJs("window.location.href = '%s'"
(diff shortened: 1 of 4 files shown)


braxtonmckee:
This looks mostly good. I rebased it and stuff seems to be working.

We lost the 'disconnected' screen when you kill the backend server. That's pretty important, since otherwise you get into a bad stale state.  Once we get back in there, this is good to go. The code is much cleaner!

darth-cheney:
@braxtonmckee I've added the close handler (one line change) and it works. Build is failing but that seems unrelated.

braxtonmckee:
Cool. Yeah, the build is quite flakey at the moment. Need to figure
something out for that. I'll review...

PR #84, opened 2019-04-08
Support callable Class objects

atzannes:
## Motivation and Context
When we define the `__call__` method of a type derived from `Class`, we want the objects of that type to be callable:
```       
class CallableClass(Class):
    def __call__(self, *args, **kwargs):
        return 42

obj = CallableClass()
self.assertEqual(obj(), 42)
```

## Approach
Implement `tp_call_concrete` for PyClassInstance objects. Its implementation conveniently calls the helper method `callMemberFunction`

## How Has This Been Tested?
I wrote a unit-test which was failing before these changes but now is passing

## Types of changes

## Checklist:


atzannes:
@braxtonmckee this is ready for review & merge

PR #86, opened 2019-04-11
Popover Bugfix

darth-cheney:
A fix for Popover rendering, which broke after changes to HTML generation.
## Motivation and Context

Since the switch to `HTMLElement` based generation in the `Popover` cell, popover actions were not working inside the browser and were throwing a very cryptic jQuery-nested error.

## Approach

The source of the problem had several aspects. The `updatePopovers()` function deals with re-rendering important information in any popover by way of internal function calls (ie, `title`, `content`, and `placement`). 
  
In the old setup, these functions looked for the inner HTML contents of nested `div`s in the popover target in order to get information about the title, content, and placement. It was the latter value -- `placement` -- that ended up being the root of the problem. Unlike the other functions in `popover` for content and title, the `placement` callback gets passed *different arguments and also changes the context of `this`* compared to the other two.
  
This was compounded by a shared `getChildProp` function that assumed that the inner HTML of nested divs with the correct information could be parsed out without any problems. The `HTMLElement` generated HTML adds whitespace and newlines in pretty-print fashion. `getChildProps` was expecting a single string ("bottom", "top", whatever) *without any whitespace*. The presence of the whitespace is what threw off the popover module.
  
This PR does the following:
1.  Puts `data-placement` as a property on the anchor target for the source popover, rather than as information in a child div
2.  Changes `updatePopovers` `placement` function to explicitly read this value by passing the proper arguments to the callback and reading the internal dataset for that `data-placement` value
3.  Moves `updatePopovers` completely into `CellHandler`
## How Has This Been Tested?
Tested using the `UninitializableService` alert popover on the front page of the basic frontend test. Also ran the full `web` test suite and everything has passed.

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/object_database/web/cells/cells.py
+++ b/object_database/web/cells/cells.py
@@ -1757,11 +1757,11 @@ class Popover(Cell):
             .set_attribute('style', self._divStyle())
             .with_children(
                 HTMLElement.a()
-                .set_attribute('container', 'body')
                 .set_attribute('href', '#popmain_%s' % self.identity)
                 .set_attribute('data-toggle', 'popover')
                 .set_attribute('data-trigger', 'focus')
                 .set_attribute('data-bind', '#pop_%s' % self.identity)
+                .set_attribute('data-placement', 'bottom')
                 .set_attribute('role', 'button')
                 .add_classes(['btn', 'btn-xs'])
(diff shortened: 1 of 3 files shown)


braxtonmckee:
looks good.

PR #87, opened 2019-04-12
Warn about unused variables and Fix those Warnings

atzannes:
## Motivation and Context
Unused variables represent either left-over dead-code or indicate that we forgot to use an input into the code coming after the variable definition. It's a good idea to get warnings about them and address them.

## Approach
Removed the `-Wno-unused-variable` compiler flag from `setup.py` and `Makefile`, recompiled the code and fixed all unused-variable warnings.

## How Has This Been Tested?
TravisCI is happy

## Types of changes

## Checklist:

### DIFF
--- a/typed_python/PyPointerToInstance.cpp
+++ b/typed_python/PyPointerToInstance.cpp
@@ -114,7 +114,6 @@ PyObject* PyPointerToInstance::pointerGet(PyObject* o, PyObject* args) {
 //static
 PyObject* PyPointerToInstance::pointerCast(PyObject* o, PyObject* args) {
     PyInstance* self_w = (PyInstance*)o;
-    PointerTo* pointerT = (PointerTo*)extractTypeFrom(o->ob_type);
 
     if (PyTuple_Size(args) != 1) {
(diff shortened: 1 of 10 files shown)


typed_python/PyPointerToInstance.cpp
```diff
@@ -114,7 +114,6 @@ PyObject* PyPointerToInstance::pointerGet(PyObject* o, PyObject* args) {
 //static
 PyObject* PyPointerToInstance::pointerCast(PyObject* o, PyObject* args) {
     PyInstance* self_w = (PyInstance*)o;
-    PointerTo* pointerT = (PointerTo*)extractTypeFrom(o->ob_type);
```
atzannes: @braxtonmckee is this variable unused because we forgot to use it? (The rest of the unused variables in this PR seem legitimate candidates for deletion, but I am unsure for this one).

braxtonmckee: I think we can get rid of it. We don't need to know the source pointer type to do a cast.

PR #89, opened 2019-04-18
Asynchronous Dropdown Cell

darth-cheney:
This PR implements an `AsyncDropdown` Cell, which is a `Dropdown` that can dynamically load the displayed contents when the dropdown is opened (as opposed to generating those contents ahead of time).
  
## Motivation and Context

Not only is it useful to be able to put arbitrary elements inside of a Dropdown, but because some Cells require computation on the back end that could take a bit of time (relative to other UI components), it's helpful to load those inner Cells dynamically when the user opens the dropdown.

### How it Works ###
We introduce two new Cell types, `AsyncDropdown` and `AsyncDropdownContent`. The latter should not be used directly, but is instead created and bound to the former at instantiation.
  
An `AsyncDropdown` cell takes a function as it's constructor argument. We expect this function to do some kind of computation, after which it returns a Cell or Cell structure that can be displayed (that's the asynchronous part). We make use of a special `Slot` that tracks whether or not the dropdown is in the open position. When this changes to open, `AsyncDropdown` will pass that information to its child `AsyncDropdownContent` component and tell it to call the passed display function (which is bound to a `Subscribed` Cell). If that passed function is working correctly, voila, the new content is displayed.
 
**Note that this implementation is a bit hacky, due to event binding difficulties outlined in the sections below**.
  
### Caveat: Front End and Event Binding Issues ###
The biggest difficulty in implementing this component was adding event binding to "know" when the dropdown, which is a Bootstrap Dropdown, had opened or closed. The problem extends from the lifecycle structure of `Cell` objects themselves, the current protocol on the websocket, and how elements are handled on the front end. As it stands, there is no uniform way to bind events to a Cell when it gets inserted to the DOM besides inlining callbacks into event handlers directly in the HTML (ie `<div onclick="<some-js-here>"></div>`.  This is not flexible enough for building complex UIs.
  
For now, we've implemented a hacky way of doing this, by adding a special method to the global `cellHandler` object on the JS side. It is defined [here](https://github.com/APrioriInvestments/nativepython/blob/eric-component-1/object_database/web/content/CellHandler.js#L211). Because we need to add listeners for the Bootstrap `show` and `hide` events but have no way of doing that naturally with Cells, we instead bind this `dropdownInitialBindFor` method to the `onclick` inline click handler for the dropdown's button element. It checks to see if it is the first time that its click event has fired. If so, it binds handlers for the Bootstrap events. If not, it does nothing. This ensures that we can receive reverse messages from the UI to the Cell whenever the dropdown opens or closes.

Another related issue dealt with knowing when to re-render specific parts of the dropdown. Unlike the existing `Dropdown` cell, we cannot just re-send HTML for the cell over the wire each time, because we care about when it opens or closes and we update child cells based upon those events. To `recalculate` each time means inserting a whole new dropdown and those always render with their menus closed initially. A mix of slot use and separating the dropdown into two constituent Cells seemed to do the trick here, once we got event binding working well.
  
This implementation is fairly complicated. A refactored Cells lifecycle, WS protocol, and DOM handling pattern could greatly improve the implementation of this component and others. We have some [working notes on this possibility available here](https://gist.github.com/darth-cheney/4b5451c8bcaf31ff1f2d5765a83c7310). Otherwise we run the risk of adding incredibly specific methods to `CellHandler` for each unique type of `Cell` we create, which can get out of control quickly.
  
### Try it Yourself ###
We have created a [`DropdownTestService`](https://github.com/APrioriInvestments/nativepython/blob/eric-component-1/object_database/service_manager/ServiceManager_test.py#L297) and added it to the basic service testing application at `frontends/object_database_webtest.py`.
  
In the example, opening the menu will show a loading display, trigger a callback that waits for 1 second and then displays the completed result.
  
### Incomplete ###
This PR is not quite complete. We need to add flexibility to allow the Dropdown to take a custom loading display cell, as well as a text label for the dropdown. For now, we wanted to start a review/discussion of the implementation before finalizing anything.

## How Has This Been Tested?
In addition to the live `DropdownTestService` mentioned above, we have added a couple of tests that ensure behavior is correct from the Cells side of things.

## Types of changes

## Checklist:

- [X] My code follows the code style of this project.

### DIFF
--- a/object_database/frontends/object_database_webtest.py
+++ b/object_database/frontends/object_database_webtest.py
@@ -24,7 +24,8 @@ from object_database.service_manager.ServiceManager_test import (
     GraphDisplayService,
     TextEditorService,
     HappyService,
-    UninitializableService
+    UninitializableService,
+    DropdownTestService
 )
 from object_database.web.ActiveWebService import (
     active_webservice_schema,
@@ -95,6 +96,10 @@ def main(argv=None):
             with database.transaction():
                 service = ServiceManager.createOrUpdateService(
                     TextEditorService, "TextEditorService", target_count=1
+                )
+            with database.transaction():
+                service = ServiceManager.createOrUpdateService(
+                    DropdownTestService, "DropdownTestService", target_count=1
                 )
             while True:
                 time.sleep(.1)
--- a/object_database/service_manager/ServiceManager_test.py
+++ b/object_database/service_manager/ServiceManager_test.py
@@ -301,7 +301,7 @@ class DropdownTestService(ServiceBase):
     @staticmethod
     def serviceDisplay(serviceObject, instance=None, objType=None, queryArgs=None):
         return Card(
-            AsyncDropdown(DropdownTestService.delayAndDisplay)
+            AsyncDropdown('Dropdown', DropdownTestService.delayAndDisplay)
         )
 
     @staticmethod
--- a/object_database/service_manager/ServiceManager_test.py
+++ b/object_database/service_manager/ServiceManager_test.py
@@ -306,7 +306,7 @@ class DropdownTestService(ServiceBase):
 
     @staticmethod
     def delayAndDisplay():
-        time.sleep(3)
+        time.sleep(1)
         return Text('NOW WE HAVE LOADED')
 
 happy = Schema("core.test.happy")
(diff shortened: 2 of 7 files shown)


darth-cheney:
@braxtonmckee for whatever reason, I cannot get the "Pull Request" travis build to restart. The failure was one of the intermittent (ie irrelevant) ones, I think

braxtonmckee:
Ok. I’ll try to get some of these flakey tests fixed. I’ll try to review as soon as I am able. Thanks.

PR #93, opened 2019-05-01
Support forward references for Alternative type (simple case)

rubidge:
## Motivation and Context
Support forward resolution (works for Bexpress example)
* could work for other Alternative forward references, by reordering generated code.

## Approach
For Alternative, support overlapping members with different types -- accessors will return a OneOf type in that situation.
Forward references are represented as a Forward type, with small adjustments to make the python forward references work in the c++ case.

## How Has This Been Tested?
Trigger test with "make cpptests".  Simple tests created for NamedTuple and Alternative direct c++ types.


rubidge:
superseded

PR #94, opened 2019-05-02
Initial Layouts Implementation

darth-cheney:
This PR is the first pass at implementing a layout system for Cells.
  
## Motivation and Context ##
Currently, all layout of Cell object as displayed in the browser involves two patterns:
* The Bootstrap CSS dependency on the client side (which has relevant classes)
* Inline styling applied to Cell objects and their HTMLElement structures in the
definition of Cell classes themselves.
  
While having client-side CSS classes is perfectly fine, manual inline styling and structuring
of Cell layouts in each individual case it not the most ideal setup. Therefore we are introducing 
several new classes that each interact with each other and Cells to form a more comprehensive
layout system.
  
## Approach ##
In order to make this system work, we have made some modifications to the base `Cell` class that update the lifecycle to better handle differend layout scenarios that may emerge. This is iscussed in more detail in the "Cells Adjustments and Lifecycle" section below.
  
### `StyleAttributes` ###
Before we can really deal with styles, we need a sane way to update, combine, and translate to string any dictionary of CSS styles. So we introduce a new object called `StyleAttributes` which has the following useful behaviors:
  
* Can print a representation of itself as an HTML inline style string, ie `style="display:none; top:0px; left:12px"`
* Can add, update, or remove specific styles from the internal dictionary
* Can be merged with an incoming dictionary of style attributes (newer entries overwrite older ones)
* Can be added together, yielding a new `StyleAttributes` that is the merge (see above) of the two
  
The `StyleAttributes` object also provides a kind of standard interface on the Python side through which any styling can pass through.
  
Additionally, it provides a kind of backward compatibility in the interim. Now all called to getters/setters like `width()` on `Cell` objects can simply read or update an internal `StyleAttributes` instance (see `baseStyles` below).
  
### `Styler`, Layouts, and Children ###
We introduce an abstract class called `Styler`, which is just an object that takes some settings and returns a configured `StyleAttributes` instance. As it is abstract, it is not intended for direct use. Instead, it requires that all subclasses implement `get_styles()` for returning an updated/confogured `StyleAttributes`.
  
There are to main kinds of Stylers for the moment: Layouts and ChildStylers.
  
Layout objects describe how a given `Cell`'s children will be laid out, given that the `recalculate` method is written according to a certain specification. This is possible because all `Cell` objects now have a property called `layout` which, by default, is set to `None`.
  
ChildStylers are objects that provide styles that describe specific attributes necessary for the children of certain layouts. **Note** that this is different from setting the `layout` property of a child `Cell`: all cells can have their own layouts while also maintaining a description of how they should be displayed in a given parent layout. For example, imagine a Grid type layout where one of the children is a `Card`. The parent Cell will have a `layout` property set to `GridLayout`, the `Card` will have custom styling from the `GridChildStyler` that describes how it should look in a `GridLayout` (ie, "take up the top four cells of the grid"), but it can also have its own `layout` set to `FlexLayout`, which ensures its children will be flex.
  
See the "Cell Adjustments and Lifecycle" section below for a more concrete explanation with examples.
  
### Layouts: Grid and Flex ###
What is a `Layout` object? Simple. It's just a specific interface that takes input and can respond with `StyleAttribute` objects that will produce the desired layout in CSS.
  
For example, `GridLayout` takes two required parameters: the number of columns and the number of rows to put in the grid. Implementors can also pass lists that are mappings for each of these, where each element in the list is either an exact measure for the width of the dimension, or an `fr` measurement that is a normalized, relative measurement comparing all widths to each other. 
  
All of this potentially horrendous complexity is handled by `GridLayout` under the hood. You never really need to worry about the CSS. Just grab the `StyleAttribute` object returned from `get_styles()` and apply it to any elements in your cell. It's as simple as that.
  
For the moment, we believe that the layouts we need in the short and medium term can be described in terms of [CSS Flexbox](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout), [CSS Grid](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout), or None/completely custom layouts. OF course, the point of the Cells layout structure is to be able to remove as much required CSS knowledge as possible from Cells implementations, though some of this is unavoidable (future work will deal with basic cases, see "Future Work" below).
  
So we have created the first two types of layouts, `GridLayout` and `FlexLayout`.
  
### Cells Adjustments and Lifecycle ###
In order to get this layout system working in a reasonable way, we made a few non-breaking adjustments to the base `Cell` class that affect its lifecycle in terms of HTML generation and styling.
  
As mentioned already, we add a `layout` property that is by default set to `None`. This allows implentors to override or redefine layouts on the fly.
  
We also have added a `baseStyles` property that is initialized to an empty `StyleAttributes` instance. This property represents the most customized level of CSS styles that the root HTMLElement of the given cell will have as its inline styling (assuming implementors do so in `recalculate`).
  
Finally, we assume that all `Cell` objects that want to make use of layouts will have a required method called `styleInParentLayout()`. This method takes no arguments, but expects that it will return a configured `StyleAttributes` object. The key idea here is that the method will generate styles for the given `Cell` based on what its parent layout is.
  
What does that mean? It means if we configure certain kinds of Cells this way, we can change their parent Cell's layout on the fly and have the children adjust immediately.
  
For example, imagine a kind of view cell called `SomeWebView` that starts with a `GridLayout`, but if the user pushes a radio button it adopts a `FlexLayout`. Now imagine we have a child `Cell` of this view that needs to adjust accordingly: occupying the right-end position in the flex layout, but spanning 2 grid spaces in the grid layout. The `styleInParentLayout()` definition for this child `Cell` might look something like:
```python
def stylesForParentLayout(self):
    if isinstance(self.parent.layout, FlexLayout):
        return StyleAttributes(**{'margin-left': 'auto'})
    elif isinstance(self.parent.layout, GridLayout):
        return GridChildStyler(row_span=2, column_span=2).get_styles()
```
Now if some event or code changes the `SomeWebView.layout` property to `GridLayout` or to `FlexLayout` this specific child will adjust accordingly.
  
#### Combining Styles ####
Because we can never completely abstract away CSS from layouts and styles, it is important to realize to things:
1. `StyleAttributes` will combine (via `__add__` (`+`) or `.append()`) in such a way that the incoming instance will overwrite any common properties that they share;
2.  Styles in a `Cell` can come from up to 3 different places: layout, layoutInParent, and baseStyles. In fact, that order is how they should be added because it preserves the correct "cascade".
  
In our examples, we have been using a custom method called `style()` on view cells that implement layout structures. Generally, this method does the following:
```python
def style(self):
    layout_styles = self.layout.get_styles()
    layout_in_parent_styles = self.styleInParentLayout()
    return layout_styles + layout_in_parent_styles + self.baseStyles
```
This will return a single `StyleAttributes` instance that is a correctly combined version of all three styles. It is consumed, for example, like so:
```python
def recalculate(self):
    self.contents = str(
        HTMLElement.div()
        .set_attribute('style', self.style().as_string()))
```
  
### Future Work ###
The classes and adjustments in the PR break nothing as it stands, only adding new functionality. Future work will involve actually putting current `Cell` objects into this layout structure.
  
One strategy we believe will be useful is to create a separate `Cell` module that holds kinds of cells called `Views`. `View` Cells are simply Cells that have certain specific or paramaterized configurations of Layout structures already implemented. We have provided examples of these in the test frontend.

## How Has This Been Tested?

We have written working tests for all new classes.
  
Additionally, we have created new `ServiceTest` modules for the demo frontend that show off the layouts:
* `FlexLayoutTestService`
* `GridLayoutTestService`
* `GridLayoutWithSidebarTestService`

  
## Types of changes

## Checklist:


dkrasner:
I believe this is now stale and out of sync with the current plans, and propose to close this PR. 

We can keep the branch around for a bit in case,

braxtonmckee:
This is part of object_database

PR #99, opened 2019-05-16
WaitFor Test

atzannes:
## Motivation and Context
ServiceManager's `waitRunning` and `waitStopped` methods had no tests, and `waitStopped` was not checking the right thing. This PR fixes that bug and adds a test for both these methods.

## Approach
`not inst.isNotActive() != inst.isActive()`. We were using the 2nd form, but the first one is the correct for the check in `waitStopped`.  

## How Has This Been Tested?
The new unit-test, tests both `waitRunning` and `waitStopped`. It passes with the fix to `waitStopped` but it would fail before this change.

## Types of changes

## Checklist:

### DIFF
--- a/object_database/service_manager/ServiceManager.py
+++ b/object_database/service_manager/ServiceManager.py
@@ -165,7 +165,7 @@ class ServiceManager(object):
                 return True
 
             instances = service_schema.ServiceInstance.lookupAll(service=service)
-            if any(inst.isActive() for inst in instances):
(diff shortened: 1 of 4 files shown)


braxtonmckee:
tests look unhappy?

atzannes:
strange. Looks unrelated to my changes. I'll re-run the tests

atzannes:
@braxtonmckee ready for review again

atzannes:
@braxtonmckee can we merge this?

PR #96, opened 2019-05-12
breaking up ActiveWebService

dkrasner:
This PR cleans up ActiveWebService, adding ActiveWebService_util.py, removing extraneous code, cleaning flow, and adding documentation.

## Motivation and Context
ActiveWebService  contained all the view related and ws communications utilities. These are now separated, cleaned up, functions living in a util file. This makes it easier to follow the logic, understand variable scope, inner/inter dependencies and continue dev. 

## Approach
Move anything related to views/html and basic ws message handling out of ActiveWebService. 

## How Has This Been Tested?
Running and testing `object_database_webtest.py`

## Types of changes
- [x] Cleanup/refactor
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

Note: this work was previously branched from `daniel-maquette` but that will not be merged in the near future, so it made sense to branch from top of `dev` and checkout (+adjust) the relevant code. It was easier to check out the final commit than to check them out separately and test accordingly. Hence, everything is in one squashed commit here.

PR #91, opened 2019-04-24
VDom (maquette) integration 1

dkrasner:
This PR is part of the ongoing process to improve both the communication (WS) protocol and client side rendering framework. The main change here is the introduction of a virtual DOM js library, [maquette](https://maquettejs.org/), with minimal changes to the current setup. 

## Motivation and Context
There are a number of [good reasons](https://www.accelebrate.com/blog/the-real-benefits-of-the-virtual-dom-in-react-js/) to use a virtual DOM setup. From our standpoint, we would like to ensure efficiency in browser rendering on DOM changes as well as a simple, lightweight, API that will be the foundation of a unified WS protocol (see [here](https://gist.github.com/darth-cheney/4b5451c8bcaf31ff1f2d5765a83c7310) for a draft)). 

Additional motivation include a simplified client-side setup to allow for better event binding, callbacks and debugging (see discussion in [this PR](https://github.com/APrioriInvestments/nativepython/pull/89)). 

[Maquette](https://github.com/APrioriInvestments/nativepython/pull/89) is very lightweight, has great documentation, and is designed to do just this sort of thing and do it well.

## Approach
All interaction with the DOM is now handled by the maquette library. All html coming over the WS is [converted](https://github.com/APrioriInvestments/nativepython/blob/daniel-maquette/object_database/web/content/CellHandler.js#L222) to a maquette vdom element and all updates, appends and replacements are handled by the maquette [projector](https://maquettejs.org/typedoc/interfaces/projector.html). 

**Note**: the general client-side rendering and logic has **not** been altered. For example, the two caches designed to deal with DOM interactions are still present: 
* [this.cells](https://github.com/APrioriInvestments/nativepython/blob/daniel-maquette/object_database/web/content/CellHandler.js#L21) (designed to keep copies of content, some rendered, some to be rendered DOM elements) and  
* [div[id="holding_pen"]](https://github.com/APrioriInvestments/nativepython/blob/daniel-maquette/object_database/web/content/page.html#L48) (designed to allow for "off-screen" DOM element manipulation). 

Neither of these is effectively changed from a logic standpoint. In addition, raw html is still coming across the wire now converted to vdom elements. The goal is to have minimal impact on anything server-side, making an incremental change, but also not avoid development which will become moot with near future changes. 

[...]

PR #102, opened 2019-05-27
Cell Components

darth-cheney:
This PR incorporates an **interim JS component structure** for Cells.
## Motivation and Context

To date, all UI-specific code has lived on the Python side inside of `Cell` definitions. This includes properties like styling and layouts as well as the structure of HTML that should be sent to the browser in order to properly render the Cell visually.
  
As a part of refactoring, we've decided that we should push as much UI functionality into the actual UI (ie the browser) as possible, while having `Cell` objects on the Python side track state changes and communicate them to front-end components as needed.
  
This PR represents an interim transition towards more front-end responsibility in the form of [basic JS "Components"](https://github.com/APrioriInvestments/nativepython/tree/daniel-components-1/object_database/web/content/components) that encapsulate the visual display of `Cell` objects.
  
To see how this fits with the wider roadmap, see "Future Work" below.

### Dual Function ###
The most important thing to know for now is that we have *kept the current Cell functionality intact* so that we can "switch" on and off whether or not a given `Cell` type uses the old-style rendering (ie the older WS protocol and element handling on the front end) or the new-style JS Components.
  
Whether or not a given `Cell` is rendered as a "Component" or old-style element is determined by the presence of the Cell name in the `availableCells` dictionary in `web/content/page.html`. In order to switch back to old-style rendering (which should work just as it had in the past), simply comment out the Cell from that dictionary.
  
### Component Structure and Lifecycle ###
All Cell components inherit from the `Component` class, which is quite minimal for the moment. We have done this intentionally -- with the current pattern it should be relatively easy to switch to any other widely-supported component framework (React, etc). DOM interaction is handled by the [maquette library](https://maquettejs.org/) and is built on top of the work in [this PR](https://github.com/APrioriInvestments/nativepython/pull/91). It also makes the code more comprehensible and reduces overall complexity.
  
For now there are two main lifecycle methods to any Cell Component:
* `render` -- This is called just before a component is rendered to the maquette v-dom. The contract is that this method should always return some `hyperscript` (which is an object-based description of an eventual virtual dom element) structure, sort of like the equivalent of setting `self.recalculate` on the Python side of things.
* `componentDidLoad` -- This is called at the end of any message handling session that has added or updated components in the browser, and tell us that the given component was, in fact, rendered to the virtual dom. We use this to do things like setup extra event listeners, check for the presence of other components, etc
  
### Python Cell Modifications and `exportData` ###
In order to facilitate comprehensible components, we added several pieces of information to the object that is sent over the WS when a Cell is to be rendered in the browser. This happens [here](https://github.com/APrioriInvestments/nativepython/blob/daniel-components-1/object_database/web/cells/cells.py#L476).
  
The main piece of information is a new dictionary called `exportData` that is now initialized for every `Cell` subclass as an empty dictionary. We use this to add pieces of information that are updated in each `self.recalculate` call, but would not otherwise have made it to the front end before -- data that is required (currently) for the HTML structure to be properly rendered to the DOM. We have updated all `Cell` definitions on the Python side to use this `exportData` dictionary as needed.
  
On the JS side, this dictionary is made available to all Components as `this.props.extraData`. You will see this throughout Component class definitions wherever it is needed.
  
The goal is to eventually obviate the need for things like this extra data once we have a more standardized state management / socket update pattern in place.
  
### Replacements ###
Because we have left the dual-function intact for this transition phase, we have not modified the process of string replacements that occurs throughout the system. This is a complex issue and requires careful modification and refactoring that is best handled in future work.
  
For now, Component classes still make use of replacements, which are now brought over the wire in their raw form. We've created a specific class called `ReplacementsHandler` that deals with a lot of the complexity of dealing with replacements in Component definitions. I will not go into too much detail here, but the important things to know are that all components have access to an internal `ReplacementsHandler` instance at `this.replacements` and that we have convenience methods for looking up replacements for a given Component type.
  
Note also that there are two types of replacements:
* `regular` -- Single keyed replacements, like `____header__`
* `enumerated` -- Two or more dimensional replacements like `____child_2__` or `____row_2_5__`
  
Of course, we also anticipate removing all of this replacement handling in future work since it should become unnecessary. For now it supports the interim transition and seems to work without issue.
  
### Which Cells Are Actually Used, Anyway? ###
In going through all Cells and creating Component versions, we realized that we aren't sure which Cells are actually used in the "real" production system. Obviously we cannot completely know. With this in mind, we have added a little micro-testing class to the client that will allow you guys to "turn on" and monitor which Cell components are actually rendered into a page.
  
This class is called `CellTracker` and you can interact with it from the browser console.
  
Here's how it works. To start "tracking" which Cells are used, simply call `CellTracker.start();` from the browser console. It will track across pages and even sessions until you tell it to stop (data is stored in `localStorage` and reloaded any time you reload the page). To stop tracking you can enter `CellTracker.stop()` into the console, and likewise to clear the stored tracking data you can enter `CellTracker.clear()`.
  
  The `CellTracker` object will have two sets, `used` and `unused` that will contain the names of different Cells. Note that this only tracks Cell types that have been turned into components.
    

## How Has This Been Tested?
Testing has happened almost exclusively via the provided frontends in the open source repositories, in particular the test frontend in `object_database/frontends`.
  
Because this is an interim/transition phase, we have opted to skip the rather large overhead of adding front-end testing (via headless/selenium or similar) and unit testing components, given that all of this will likely be subject to change sooner rather than later.
  
The exception to this is the `ReplacementsHandler` class, which does some fairly complex things and had to be developed using TDD. Those tests are available [here](https://github.com/APrioriInvestments/nativepython/blob/daniel-components-1/object_database/web/content/components/tests/replacements-tests.js). The tests were written in [mocha](https://mochajs.org/) and [chai](https://www.chaijs.com/) and require a node environment with correct dependencies (which we've kept out of the repository for now). I can provide separate instructions for running these tests by request if you wish.

## Merging ##
The initial work here starts from the head of the `daniel-aws-clean` branch and [subsequent PR](https://github.com/APrioriInvestments/nativepython/pull/96), as the breakup of `ActiveWebService` and related cleanup allows for easier debugging. Although technically the same in the end, it might make sense to consider merging that first, to keep the review and discussion logically separated. 

Regardless of which way we proceed it will be important to thoroughly test both stacks of changes in the _true_ production environment. 

### Bundling ###
In the immediate future, and once this PR is accepted/merged, we will incorporate a [Webpack](https://webpack.js.org/) bundling flow to the front-end. This will allow us to have sane imports and dependencies across module files for the client side, and will make writing tests and future changes much easier. We anticipate this process should only take an afternoon at most.
  
### Cells and Websocket ###
Once this PR has been tested in the live environment, tested, and merged, we will move as quickly as possible to dropping "dual functionality" and use pure components. This will also go hand-in-hand with slimming down what `Cell` objects to on the Python side, how we communicate over the websocket, and how state changes are managed, monitored, and updated.
  
The end goal is to have Python Cell objects almost exclusively hold and broadcast state changes to matching components that are alive on the browser side.

## Types of changes

## Checklist:


darth-cheney:
These changes were rebased and manually merged in https://github.com/APrioriInvestments/nativepython/commit/ad238ce37e9482c8a34f571254115678e7745fcf

PR #103, opened 2019-05-29
Code Coverage with codecov

atzannes:
## Motivation and Context
Track how well our tests cover our code with `coverage` and the service `codecov`. This PR just sets up the integration to get the ball rolling, but it there will be several follow-ups needed:
- Our coverage for tests that uses subprocesses is unexpectedly low. This is because we have to take care to start the sub-processes in a particular way (which I'm not yet familiar with) so that they too will record coverage info and aggregate it back. We need to figure out how that is done separately.
- It will be nice to have a coverage badge on our README to show off our good coverage, when our coverage becomes better. I think that the above bullet is the first thing to figure out to increase our coverage, then we can think about increasing it by writing more tests.
- codecov has support for C++. Figure out how that applies to the C++ parts of our code
- write more tests :)

## Approach
Run the TravisCI tests through `coverage` to generate the `.coverage` data, and send it to `codecov` for historical tracking and displaying.

## How Has This Been Tested?
Checked that we get coverage info correctly transmitted to codecov. It looks like the data pipeline from TravisCI to codecov is working properly. 

## Types of changes

## Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,4 +11,5 @@ nativepython.egg-info
 /testcert.cert
 /.venv
 /.eggs
-
+/.coverage
+/.tox
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ env:
 
 install:
   - pip install pipenv
+  - pip install codecov
   - pipenv install --dev --deploy
   - sudo apt-get install -y gdb  # install gdb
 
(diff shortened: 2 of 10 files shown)


atzannes:
@braxtonmckee  ready for you to take a look

PR #106, opened 2019-06-21
Component Webpack and Tests

darth-cheney:
This PR implements [Webpack](https://webpack.js.org/) component bundling and a testing environment/framework for all JS components and utilities.

Please see the "Important Info" section below.
  
## Motivation and Context

In order to provide automated testing for Components and their utilities, as well as having a manageable module structure, we need to be able to use ES6 imports easily. Webpack essentially forms a tree of all import/export dependencies and "bundles" them into a single JS file, which is then required by the `page.html`file.

We need a way to manage modules and also to easily test components.

### Node Environment / Makefile Additions ###
In order to have both the testing and webpack abilities at the ready, we needed to add a node environment and dependency installation process to the Makefile. We use `nodeenv`, which is very friendly with `virtualenv`. The installation process is seamless.
  
All build or test commands can be run manually from `object_database/web/content`, since that is where the main `package.json` file resides.

The Makefile has also been adjusted to run the JS test suite automatically when using `make test`.
  
#### Building  ####
In order to produce the correct JS dependencies, you need to build the bundle. Note that we do not include this bundle in the repository, so if you try to run the test frontend without first building you will see an import error in the browser console.
  
To run the build process, be sure you are sourced into the venv created by the install and navigate to `object_database/web/content`. Then run the command `npm run build`. That's it. Now you should be able to run the frontend as normal.
  
#### Testing Manually ####
To run just the JS tests (without using the Makefile), be sure you are sourced into the venv and navigate to `object_database/web/content`. Then run the command `npm test`. That will run the full test suite.
  
### Testing Framework ###
For testing in JS, we use a combination of [Mocha](https://mochajs.org/) and [Chai](https://www.chaijs.com/). These are well supported and pretty standard in the JS world.

## How Has This Been Tested?
We ran the full test suite, including the JS tests.
  
See the "Important Info" section below for additional information. We used the test frontend to ensure that components were in working order.

## Important Info
The work implemented in this PR is not new. However, the `dev` branch and our working branches got so far off base from each other that this PR branch is in fact a rebase of several branches on top of `dev`. This means that there might be some weirdness that we have not caught, since the resolution process required a lot of manual work. PLEASE BE SURE TO TEST THIS IN A LIVE UI.

## Types of changes

## Checklist:


braxtonmckee:
Cool.  (a) Does the webpack get run when we return setup.py and get
included in the built binary? (b) what's the best dev process when you're
iterating on the JS in this model?

darth-cheney:
@braxtonmckee Nothing webpack related should be included in the setup.py binary. These changes will only affect files in `object_database/web/content`.
  
The dev process is pretty simple: you just need to run the build process whenever you make changes to any JS files. We can certainly set up a watch process down the road and add that. For now it's easy enough to just run the build (the procedures for doing that are outlined in the PR). Does that answer your question?

braxtonmckee:
Ok, so on setup.py, when we deploy this how will the files in
object_database/web/content get populated?

dkrasner:
@braxtonmckee we can commit the latest js bundle build with every PR - that way you can just import and use it (granted that your prod setup is similar to the web test we use/see)

braxtonmckee:
The prod setup is slightly different. We install the python library, and
expect to be able to run the active web service out of it. Ideally setup.py
would build the webpack and include its files in the manifest. It already
does this with page.html and the other javascript files.

dkrasner:
@braxtonmckee the main bundle is added, and included in the MANIFEST.in file, see [this commit](https://github.com/APrioriInvestments/nativepython/pull/106/commits/4e42b52e995ad9f11964bea16ce3251a4f743a4c)

That seems better than having setup.py install the entire node env, webpack etc just to build this bundle 1x

Issue #128, opened 2019-06-25
Frontend should be able to set the current URL


braxtonmckee:
Right now, when we click a button and navigate to a URL provided for an object-database page, we actually have to reload the page. We should be navigating internally to that page (so we retain our session state) and setting the url in the browser, but not bouncing the page.

PR #130, opened 2019-06-25
Eric named children update

darth-cheney:
This PR represents a stream of work on Cells/Components that future-proofs them for a transition away from replacements.

## Motivation and Context

This stream of work (several branches that had long ago diverged from today's updated base `dev`) contains crucial changes to both Cells and Component structures that will allow us to do the following:
* Maintain backward compatibility with the current replacements setup
* Include potential functionality for a rendering process that doesn't use replacements
* Provide initial, non-breaking functionality (described below) that will allow us to easily transition to a new rendering process

## Approach
We introduce three new concepts in this PR: `namedChildren`, property validation, and an "initial structure" test route.
  
### `namedChildren` ###
If we are going to transition away from arbitrary string replacements (ie `____contents__` or `____data_0_2__`), the front end components will still need to know which children to render in which part of their own structures. In other words, Components need to be aware of their own children *by name*. Therefore we have introduced a new dictionary on all python Cell objects called `namedChildren` and updated every single Cell in the file to map children to a specific name. 
  
Note that we have NOT removed the regular `children` dict that uses replacements -- instead these are used side-by-side for the moment for backward compatibility purposes.
  
On the JS/Component side, every component "knows" if it is using replacements or not. The new pattern is, when describing component structures that have any children at all, to test if the component is currently using replacements (for this PR the answer will always be true by default) and, if so, use the already-existing framework for replacements. If it is not, then we use a separate pattern for named children.
  
All new Cell/Components that get created will need to account for both of these cases until such time that a final transition away from replacements is made.
  
For the moment, you should notice **no changes in the current UI** as a result of these additional features, which are effectively "turned off" for now.
  
### PropTypes Validation ###
As our JS Components mature, we want to have a better handle of the types of properties that are passed to them. Currently this happens via `Cell.exportData` that is passed into each individual component constructor. The new `PropertiesValidator` will validate these props whenever a new component is created, so long as the component describes a static, class-level object called `propTypes`.
  
This `propTypes` object also serves the purpose of documenting -- in code, no less -- what properties a given component can expect from the Cell on the python side.
  
PropertiesValidator was created using a test suite that runs as a part of all JS tests.

### Initial Structure ###
The base `Cell` class has been given a couple of new methods that allow it to describe itself -- and its known children -- recursively as JSON objects. There are two main Cell methods to know about: `currentStructureFromCell` and `getChildrenJSON`.
  
#### `currentStructureFromCell` ####
Given a cell object as an argument, this method will return a dictionary that can be serialized to JSON which describes (resursively) the structure of the passed cell and does the same for any children of that cell, all the way down the chain. This method will work on any cell, allowing us to get just subsets of the full Cell hierarchy if needed. But in practice it will probably be used on initial structure calls (see below) and the `RootCell`.
  
#### `getChildrenJSON` ####
This is the recursive function consumed by `currentStructureFromCell` that will go through a given Cell tree/subtree's children and respond with a dict structure that can be serialized to JSON.
  
#### Initial Structure Request ####
In `ActiveWebService` we have added a route (`/api/<service-name>/initial-structure`) that will respond with a JSON representation of the RootCell of the given service. This route is not used anywhere else in the application yet, and is only present for two purposes: 1) to help with current debugging; 2) to better prepare for the dev work of deprecating replacements as a rendering technique.
  
If you have a frontend running, you can see an example by simply [clicking this link](http://localhost:8000/api/services/initial-structure)
  
Note that we do not ever plan on sending this information over the socket -- the initial structures are too large and best left as an initial HTTP call made by the client when first loading any page. That is the implementation plan going forward.
  
For now, **you should not notice any changes in the current UI** as this is unused in the environment for the moment.

## How Has This Been Tested?

The new properties validation was developed using TDD and its tests are available as a part of the full JS-side test suite.
  
Other changes were tested in the demo frontend and by running the usual cells tests.

## Types of changes

## Checklist:


object_database/web/cells/cells.py
```diff
@@ -1811,19 +1891,20 @@ def getChild(self):
...
+            self.children = {"____child__": childCell}
+            self.namedChildren['child'] = childCell
 
 
 class Subscribed(Cell):
     def __init__(self, f):
         super().__init__()
 
-        self.f = f
+        self.f = f  # What is f? A lambda?
```
braxtonmckee, line 1903: yes, it's a lambda, and we should rename it. IT's the function that generates the actual cell.

object_database/web/cells/cells.py
```diff
@@ -1863,6 +1947,11 @@ def recalculate(self):
 
 
 class SubscribedSequence(Cell):
+    # TODO: Get a better idea of what is actually happening
+    # in this cell. For example, what is with all the existing
+    # items funging, and what is actually needed in terms of
+    # information to display this correctly in the new
+    # JS-component based setup?
     def __init__(self, itemsFun, rendererFun, asColumns=False):
```
braxtonmckee, line 1955: This thing calls 'itemsFun' to get a list of items and then 'rendererFun' to produce the actual cell contents for each 'item'. The 'item' is really just a way of tracking the unique things in the sequence. Three things can happen: new items, items deleted, and items reordered.

object_database/web/cells/cells.py
```diff
@@ -1958,14 +2053,25 @@ def recalculate(self):
 
 
 class Popover(Cell):
+    # TODO: Does title actually need to be a cell here? What about detail?
+    # What is the purpose of the sortAs method here and why are we using
+    # it on the title cell?
```
braxtonmckee, line 2058: SortAs determines how any particular cell would sort in a Table. We have to pick something...

object_database/web/cells/cells.py
```diff
@@ -2016,6 +2122,9 @@ def sortsAs(self):
 
 
 class Grid(Cell):
+    # TODO: Do the individual data cells (in grid terms) need to be actual Cell objects?
+    # Is there a way to let the Components on the front end handle the updating of the
+    # data that gets presented, without having to wrap each datum in a Cell object?
```
braxtonmckee, line 2127: Yes it needs to be a cell, because we almost always render a 'Subscribed' in a grid so that the values can tick. I can imagine an optimization where we would detect if the rendererFun returned a python primitive like text or something, in which case we could dispense with sending a cell.

object_database/web/cells/cells.py
```diff
@@ -2547,7 +2710,7 @@ def recalculate(self):
 class Clickable(Cell):
     def __init__(self, content, f, makeBold=False, makeUnderling=False):
         super().__init__()
-        self.f = f
+        self.f = f  # What is this?
```
braxtonmckee, line 2713: it should be called 'onClick'.  It performs the button action when clicked. If it returns a string, then we navigate to the specified url. See #128 for more on this, BTW.

object_database/web/cells/cells.py
```diff
@@ -2811,13 +2988,18 @@ def recalculate(self):
             self.exportData['minLines'] = self.minLines
         self.exportData['keybindings'] = [k for k in self.keybindings.keys()]
 
+        print()
```
braxtonmckee: can we get rid of this

object_database/web/content/components/Expands.js
```diff
@@ -2,6 +2,15 @@
...
  */
 
+/** TODO/NOTE: It appears that the open/closed
+    State for this component could simply be passed
+    with the Cell data, along with what to display
+    in either case. This would be how it is normally
+    done in large web applications.
+    Consider refactoring both here and on the Cells
+    side
+**/
```
braxtonmckee, line 12: given what we're doing, we should probably make a richer set of cells (e.g. a TreeView cell) instead of trying to cobble them together with individual pieces on the python side.

object_database/web/content/components/Grid.js
```diff
@@ -70,12 +98,34 @@ class Grid extends Component {
...
                 );
             });
-        } else {
-            return []
         }
     }
 
-    _makeHeaderElements(){
+    _makeReplacementRowElements(){
+        return this.getReplacementElementsFor('child').map((row, rowIdx) => {
```
braxtonmckee, line 105: not sure if this pertains the same in this model, but you can have grids with no children, and that's not supposed to throw an exception.

braxtonmckee (commented):
this looks good. See my comments below. I'll try it out before merging it.

object_database/web/cells/cells.py
darth-cheney: Yup. I'll remove and push.

object_database/web/content/components/Grid.js
darth-cheney: I see. This should be a one line change. Let me try it out here and if it works I'll push.

PR #131, opened 2019-06-26
Initial SplitView Cell

darth-cheney:
This PR represents an initial implementation of a SplitView cell.

## Motivation and Context

We need a way to easily (and declaratively) compose views and subviews in the browser that divide up space on the screen. SplitView is a kind of "container" component that wraps child cells in this manner.

## Approach

**Note**: This PR is an initial working version that opens up the conversation about the API. We can adjust the API in the process of reviewing this PR.
  
### Behaviors ###
The `SplitView` cell has the following behaviors:
* It will take up 100% of the width and height of its parent element in the browser -- ie, it will be "greedy" and take up as much space in both directions as it can.
* It will display its child cells in either the horizontal ("row") or vertical ("column") direction. The default is "row".
* Child cells are specified as tuples, where the first element is the child cell instance and the second is a ratio describing how much space, compared to the smallest member, the given cell should try to gobble up in the row or column
* SplitViews are composable and nestable. You can have a SplitView of SplitViews of SplitViews.
  
### Example ###
The following will create a horizontal (row) SplitView where the second child cell
will be in an area that takes up twice the amount of space as the first cell area:
```python
SplitView([
    (Card(Text("First Element")), 1),
    (Card(Text("First Element")), 2)
])
```
The following will do the same, but in the vertical direction:
```python
SplitView([
    (Card(Text("First Element")), 1),
    (Card(Text("First Element")), 2)
], direction="column")
```

### Playground ###
This branch also includes the initial version of the Playground (which we are working on making updates to elsewhere). We have included several Playground examples for SplitView.

## How Has This Been Tested?

We are using the Playground to test these visually.

## Types of changes

## Checklist:


braxtonmckee:
This is cool! But two things: (a) I think row and column somehow are swapped? (b) can we make it resizeable or is that something that should be saved for another PR?

darth-cheney:
@braxtonmckee I can check to see if these were swapped. I might have mislabelled things in the playground but the actual display values should be accurate ("row" and "column" map 1:1 with the underlying flexbox values).
  
As for the resizing, what do you mean exactly? Using mouse events?

dkrasner:
if flex `row` implies that the children are laid out in a single row, and `column` that they are stacked in a single column 

this might seem counterintuitive because it looks like you are making rows when stacking children but since it's the parent div that is flex'ing in a given direction, not the children organizing themselves, it makes more sense

we could flip it but i think that could lead to confusion down the line

braxtonmckee:
OK, but the fact that this is flex is an implementation detail. From a user
standpoint, I want to be able to specify that we are going to _get_ columns
or rows. If you think that's going to be confusing terminology, I might say
'vertical split' for when you have them left-to-right and 'horizontal
split' when you have one above the other.

braxtonmckee:
yeah, on resizing, it would be nice if I could pass a flag to make the
splitter draggable using the mouse. This might be a separate component,
because if we're going to do that we'd want some styling so that the
separate panels and the drag pane are explicitly visible.

darth-cheney:
@braxtonmckee On the matter of convention about which axis things "split" on -- not a big deal. We just swap the terms and I can do that however you wish.
  
Regarding the resizing -- probably best to do this in a different component. But while we are at it maybe let's agree upon and tighten up what this component here does. For example, I'm just realizing that you wanted this component to only display two children and split on one axis or another. That is not exactly what is going on in this PR. Here we can have any number of children split down the given axis. 
  
Would you like me to update this to be more strict -- ie, just take two children? If so, we can use that as a basis for a component that will have a visible "divider" along the split axis and that (with some effort) will be able to resize the two areas with the mouse.
  
Otherwise I say keep this splitter and we make a second more explicit one that handles just two areas and can be dynamically resized.

braxtonmckee:
I think it's only necessary to just have two children if you have a
resizer. If this is static, great to have as many as you like.

dkrasner:
@braxtonmckee as another example of SplitView, take a look at playground on [this branch](https://github.com/APrioriInvestments/nativepython/tree/daniel-playground-refact)

i spent a little time and rewrote it using SplitView as the basic layout cell/component. It's not quite done but you get the idea

darth-cheney:
I'm not exactly sure how to proceed here. Should we make a separate `ResizableSplitView` in a separate PR?

braxtonmckee:
Yes. I think we should separately make a `ResizablePanel`. We might want to see if there's an existing UX nomenclature we can borrow from for some of these words.  I was thinking of trying to distinguish between pure layout cells (like the split view, that don't have any UX themselves) and components that have some visible state and tooling (like a resizeable splitview)  by calling one a 'View' (or maybe it should be called a `Layout`) and calling layout-ish things that have state `Panel`s.

darth-cheney:
@braxtonmckee Cool. Let's go with `View` for wrappers that mostly describe element layouts but not necessarily interaction, and `Panel` for interactive kinds of Views. We can work on the ResizablePanel separately.
  
Is there anything else you would like to change or is this PR ok to merge in?

PR #136, opened 2019-06-27
WIP: record coverage for subprocesses

atzannes:
## Types of changes

## Checklist:


object_database/service_manager/SubprocessServiceManager.py
```diff
@@ -109,7 +108,6 @@ def startServiceWorker(self, service, instanceIdentity):
                         self.authToken,
                         '--log-level', self.logLevelName
                     ],
-                    cwd=self.storageDir,
```
atzannes: @braxton do you remember why we have this here? The docs say it's useful for finding the executable if the first arg of Popen is a relative path, but we are clearly not using it for that. Is it so that any output produced by the sub-process exists in the tmpdir and we can clean it up at the end?

atzannes:
closing PR to prevent double builds while I work on it

PR #145, opened 2019-07-02
Multi-process coverage and codecov configuration

atzannes:
## Motivation, Context, and Approach
For coverage to correctly track lines of code visited by different processes started with `subprocess.Popen`, we needed to make several configuration changes, and some code changes. Notably, we cannot call `os._exit(errno)` because [it prevents `atexit` from getting called](https://docs.python.org/2/library/atexit.html), which `coverage` relies on in order to produce its output. That resulted in having to revisit how we tear down the SubprocessServiceManager and its worker processes: first try to terminate (SIGTERM) and after a predetermined timeout, send the more brutal KILL signal.

Furthermore, we didn't like that codecov would mark builds as red when the total (project) coverage decreased by a very small percentage. The `project: threshold: 0.5` configuration means that a total project coverage reduction of less than 0.5% will not necessarily be marked as red.

## How Has This Been Tested?
- TravisCI tests still passing
- Coverage shows `service_manager.py` being covered
- Coverage fails to show `service_entrypoint.py` being covered. This will have to get fixed later.

## Types of changes
- Better tracking of coverage
- Better decision making whether the coverage for a build is red or green


braxtonmckee (commented):
I'm not sure this is going to work. Those hard exits are there for a reason: our C++ extension is not able to shutdown cleanly when there are running daemon threads because we start deallocating C++ structures while threads are using them and then we just segfault (which produces a nonzero exit code and can break a bunch of stuff).  If we want to do it this way, we need to ensure that we actually clean all the threads up correctly.

The other option would be to leave the `os._exit` calls in but find a way to explicitly trigger codecof's dump routines and then retain the hard os._exit.

atzannes:
Hm, that sounds like a fair amount of work for a small benefit. How about I switch back to using `os._exit` and get the rest of this PR merged and I can circle around later to get coverage to work? @braxtonmckee

braxtonmckee:
why don't you try manually invoking sys.exitfunc and then hard os._exit.
The problem we have is when the c++ module gets unlinked, not when python
is running.  If sys.exitfunc runs cleanup and then hands control back to
you, maybe that will just work. If that doesn't work, then what part of
this change do you want to keep?

atzannes:
@braxtonmckee looked into your suggestion and `sys.exitfunc` was deprecated in python 2.4 and does not seem to exist in python 3. It has been replaced by the `atexit` module, which does not have a public method to call the registered functions. As a possible work-around, I am keeping track of the return value of `main` (`errno`) in `service_manager.py` and `service_entrypoint.py`, I try to call `sys.exit(errno)`; if that call throws an exception I call `os._exit(errno)` so that the correct error number is set. What I don't know is whether `sys.exit` throws an exception in the cases you mentioned, so not sure this approach will work. It would be great to have a test that fails with `sys.exit()` but not with `os._exit()`.

An other option could be to try running `atexit._run_exitfuncs` before calling `os._exit()`, but if the clean-up code in atexit could be causing the segfault, we're back at square one.

A final option, which would be quite hacky, would be to try and figure out what function `coverage` is registering with `atexit` and call that directly right before calling `os._exit()`, but that could break when we update the `coverage` package, because it is not part of its public API.

Thoughts?

braxtonmckee:
I merged changes in that should prevent us from crashing at process exit by making sure we never delete Type objects. So at this point, you should be able to rebase this PR and remove the os._exit calls I used to have (I did this in mainline so you'll probably just have to deal with the merge conflicts) and it should just work without all the hacking around.

PR #146, opened 2019-07-02
Fix showing error when type passed into refcount is wrong.

szymonlipinski:
## Motivation and Context
There is a segmentation fault when argument passed to the refcount function is string or int.

## Approach
The change was to convert the argument to a string version, like it's expected by the `%S` format flag.

## How Has This Been Tested?
There is a new test which fails without the change.

### DIFF
--- a/typed_python/_types.cpp
+++ b/typed_python/_types.cpp
@@ -656,8 +656,8 @@ PyObject *refcount(PyObject* nullValue, PyObject* args) {
             )) {
         PyErr_Format(
             PyExc_TypeError,
-            "first argument to refcount %S not a permitted Type",
-            a1
+            "first argument to refcount '%S' not a permitted Type",
+            (PyObject*)a1
             );
         return NULL;
     }
--- a/typed_python/types_test.py
+++ b/typed_python/types_test.py
@@ -171,6 +171,13 @@ class RandomValueProducer:
 
 class NativeTypesTests(unittest.TestCase):
 
+    def test_refcount_bug_with_simple_string(self):
+        with self.assertRaisesRegex(TypeError, "^first argument to refcount '111' not a permitted Type$"):
+            _types.refcount(111)
+
+        with self.assertRaisesRegex(TypeError, "^first argument to refcount 'aa' not a permitted Type$"):
+            _types.refcount('aa')
+
     def check_expected_performance(self, elapsed, expected=1.0):
         if os.environ.get('TRAVIS_CI', None) is not None:
             expected = 2 * expected


typed_python/_types.cpp
```diff
@@ -656,8 +656,8 @@ PyObject *refcount(PyObject* nullValue, PyObject* args) {
             )) {
         PyErr_Format(
             PyExc_TypeError,
-            "first argument to refcount %S not a permitted Type",
-            a1
+            "first argument to refcount '%S' not a permitted Type",
+            PyObject_Str(a1)
```
braxtonmckee: So, two points. First, this will leak a PyObject.  Pretty much any function that produces a new PyObject needs a corresponding PyDecref operation to balance it. 

Secondly, I looked at the docs and %S was the right code after all. The problem is that this function is like a C-style printf - there's no type information so it just passes the argument in direclty, and we're passing a PyObjectHolder which doesn't know how to convert itself to the underlying PyObject.  So I think the right thing to do is to put it back to `%S` and to explicitly cast `a1` to a PyObject*...

typed_python/types_test.py
```diff
@@ -171,6 +171,18 @@ def pickRandomly(self):
 
 class NativeTypesTests(unittest.TestCase):
 
+    def test_refcount_bug_with_simple_string(self):
+
```
braxtonmckee: no extra line here unless you have a comment or docstring please.

typed_python/types_test.py
```diff
@@ -171,6 +171,18 @@ def pickRandomly(self):
...
+        with self.assertRaises(TypeError) as context:
```
braxtonmckee: there is something called `assertRaisesRegex` that lets you simultaneously check the type of exception and also checks that a regex matches the internals.

typed_python/types_test.py
```diff
@@ -171,6 +171,18 @@ def pickRandomly(self):
...
+            _types.refcount(111)
+
+        self.assertTrue(str(context.exception) == "first argument to refcount '111' not a permitted Type", str(context.exception))
```
braxtonmckee: Normally I'd use `self.assertEqual` which produces a nice error message if they are not equal automatically.

braxtonmckee (changes requested):

typed_python/_types.cpp
szymonlipinski: thank you

braxtonmckee:
This looks good. Pls fix the linting error and squash to one change and I'll merge. Thx.

PR #147, opened 2019-07-03
Resizable Two Element Panel

darth-cheney:
Implements ResizablePanel cell that can display two child cells in a kind of SplitView that has
a resizer along a specified (horizontal or vertical) axis.

## Motivation and Context

We needed this element to more easily adjust the width and/or height of nested views within the UI.

## Approach

Because of the complexity associated with resizing in the DOM (it requires global listeners to come and go) and the way we update elements, we decided to use the [Split.js](https://github.com/nathancahill/split/tree/master/packages/splitjs#api) library to handle the actual resizing.
  
Aside from this the implementation is straightforward: there is a new Cell called `ResizablePanel` and a corresponding JS component on the frontend. 

## How Has This Been Tested?

These were tested using the playground and appear to be working.
  
Note that in the playground examples, we are using Card cells. Cards are "greedy" along the horizontal axis (meaning they will fill available space) but are not so along the vertical axis, therefore when examining the vertical example you will notice that cards to not take up the full height of their respective panes. This is a matter of styling the Card, not the ResizablePanel or its areas.

## Types of changes

## Checklist:


braxtonmckee:
This looks good! Can you rebase and squash?

darth-cheney:
@braxtonmckee Ok rebased and squashed. Seems to work on my end.

Issue #112, opened 2019-06-24
Create a cell that can grab keystrokes and call events on the server


braxtonmckee:
We should be able to accept keystrokes and decide whether to handle them. We should be able to specify an integer priority so that higher priority handlers get to decide first if they want the keystroke.

darth-cheney:
In theory this shouldn't be a problem, but there are a few things we'll want to consider. What does it mean for an arbitrary Cell to listen for keystrokes? On the DOM side, only certain elements can actively have focus (which is required to fire key events). However we can bind key event listeners to the entire body. In that case we need to carefully manage adding/removing handlers as cells come and go.

braxtonmckee:
Yes that's a good point. I was thinking we'd have a single listener on the
DOM, send a message back to a single global cell, and do the routing on the
server side for precisely this reason.  Another thing to consider is how
does this interact with something like the CodeEditor? I think it's
probably OK for native JS components to be in front and have this be the
fallback for any non-consumed keystrokes.

darth-cheney:
> Another thing to consider is how does this interact with something like the CodeEditor? I think it's probably OK for native JS components to be in front and have this be the fallback for any non-consumed keystrokes.
  
The event model in the DOM is that events "bubble up" from the nested children to their ancestors. This is true for most types of events. Any listener can call `event.stopPropagation()` to break the bubbling. But if we have some events listening at the top ancestor level (ie the body) then we just make sure that something like CodeEditor will stop keystroke propagation when it has the focus.

darth-cheney:
Can you provide an example of how you'd envision this Cell being used in practice?
  
I am asking because I have the frontend all set up for it, along with a very basic example Cell. My problem now is that I don't really know how to communicate changes to other cells using purely the declarative structure.
  
Right now I have a `KeyAction` cell whose constructor looks like this:
```
KeyAction('Alt+t', someFunction, ['timeStamp'])
```
Where:
* `'Alt+t'` is the keyCombo that will be registered for the cell on the frontend;
* `someFunction` is a callback that will receive the data from the frontend about the event (this is handled by KeyAction's onMessage);
* `['timeStamp']` is an example of an array of keys that the action should look up on the frontend's event object and return to the server. In this example, all JS events have a timeStamp and that is data that would get sent back to Python when the key event is triggered.
  
I am not quite sure how to make an example in the playground given this setup, because it's unclear to me how to get the `someFunction` to change something in a declarative Cell structure.
  
Any ideas?

[...]

PR #152, opened 2019-07-05
KeyAction Cell

darth-cheney:
This PR implements a KeyAction Cell, Component, and front-end infrastructure for binding key combo events in the UI.

## Motivation and Context

Issue #112 identified the need to a Cell that could bind key combinations to some action.
  
*NOTE: Together, let's make some examples that are more real-world in the playground just to test this out*

### Cells / Backend ###
On the backend, we introduce a new cell class called `KeyAction`, whose constructor looks like this:
```python
def __init__(self, keyCmd, callback, wantedInfo=None, priority=4, stopsPropagation=False):
```
where a given `keyCmd` is the string representation of either a single key (`"X"`) or a modifier key and other key combination (`"Alt+i", "Meta+D", "Ctrl+Y", etc`). In the UI, when this key command is pressed, an event will occur and data about it will be sent back to the host `KeyAction` cell over the socket. `KeyAction` will then pass that information to any `callback` parameter that was provided to it, either a lambda or other function that takes a dictionary as the sole arg.
  
#### `KeyListener` ####
We introduce a new class / singleton called `KeyListener` that knows how to bind these key combinations and add them to the global event listening space in the browser. Because of the way key events work on elements, this needs to be done at the highest scoped level.
  
`KeyListener` has two useful methods: `start()` which binds a single global listener to all `keydown` events and then re-routes all constituent key combo listeners, and `pause()` which removes that single global listener. Together these can be used to start and stop listening for *any* cell-produced key events as needed. By default, we `start` once the dom content has loaded (see the `main.js` file).
  
In `main.js`, we create a singleton of `KeyListener` and then add it as a class-level property on `Component`, such that all components can access it as needed via `someComponent.constructor.keyListener`. This is, in fact, how the corresponding `KeyAction` component works.
  
#### `KeyAction` Component ####
The component `KeyAction` is a simple wrapper for the Cell data. It uses the `constructor.keyListener` method described above to bind a keybinding/listener based on the properties specified by the corresponding Cell.

### On Non-Displayedness ###
Because `KeyAction` does not have any visual component on its own, it does not create any visible UI structure on the front end. This means we have created a new "kind" of cell -- a so called "non-display" Cell, that the UI needs to know to skip over when it comes to trying to make DOM elements in any cell creation handler.
  
In order to facilitate this, the parent `Cell` class was given a new property called `shouldDisplay` which is by default set to `True`. This information is now passed over the socket on any `updateMessageFor` composed message, and the front end then has all the info it needs.

## How Has This Been Tested?

We have created a simple test in the Playground that will display a timestamp of when the `Alt+t` keycomo gets pressed.

## Types of changes

## Closes ##
#112 

## Checklist:

### DIFF
--- a/object_database/web/cells/non_display/key_action.py
+++ b/object_database/web/cells/non_display/key_action.py
@@ -1,0 +1,79 @@
+"""KeyAction Cell
+
+Listens key-combo press events from the UI"""
+from ..cells import Cell
+
+"""A class for adding Key combination event listeners
+
+to the frontend UI.
+
+Notes
+-----
+There are two types of keys you can register.
+(1) Single alphanumeric and typed symbolic keys,
+    such as:
+        `i`, `2`, `K`, `?`, etc
+(2) Modifier key plus (1) key values pressed
+    together, like pressing `Control` and `K`, etc.
+
+The possible modifier keys are:
+    `Alt`, `Control`, `Meta`, and `Shift`
+
+The way to register a `keyCmd` in the constructor for this
+class is to provide just the single character corresponding
+to the key, or, if it's a combo, to format the combo like:
+    `<modifier>+<alphanum/symbol-character>`
+For example:
+    `Alt+1`, `Control+i`, `Meta+?` etc
+
+Note also that `Ctrl` and `Control` will both work for
+the control modifier key.
+
+"""
+
+
+class KeyAction(Cell):
+    """
+    Parameters
+    __________
+    keyCmd str: A string of a key combo like `Alt+i`, `Meta+D`
+                or event single keys like `X`.
+    callback func: A function that will be called with the
+                event response dictionary data as the sole arg.
+    wantedInfo list of str: A list of keys on the UI keydown event object
+                whose values should be sent back to the Cell in the response
+                message. Optional; defaults to None.
+    priority int: The priority level of the event response.
+    stopsPropagation bool: Whether or not this KeyAction should fire and then
+                stop other KeyActions with the same keyCmd from firing
+
+    Notes
+    -----
+    For an explanation of possible `keyCmd` values, see the class comment.
+
+    For possible `wantedInfo` key names (which map to JS event data on the
+    front end), see the MDN docs here:
+    https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent
+    """
+    def __init__(self, keyCmd, callback, wantedInfo=None, priority=4, stopsPropagation=False):
+        super().__init__()
+        self.shouldDisplay = False
+        self.keyCmd = keyCmd
+        self.callback = callback
+        self.wantedInfo = wantedInfo
+        self.priority = priority
+        self.stopsPropagation = stopsPropagation
+
+        assert 1 <= self.priority <= 4
+
+    def recalculate(self):
+        self.exportData = {
+            'keyCombo': self.keyCmd,
+            'wantedEventKeys': self.wantedInfo,
+            'priority': self.priority,
+            'stopsPropagation': self.stopsPropagation
+        }
+
+    def onMessage(self, messageFrame):
+        messageFrame['data']['keyCmd'] = self.keyCmd
+        self.callback(messageFrame['data'])
--- a/object_database/web/cells/__init__.py
+++ b/object_database/web/cells/__init__.py
@@ -62,6 +62,8 @@ from object_database.web.cells.cells import (
 
 from object_database.web.cells.views.split_view import SplitView
 
+from .non_display.key_action import KeyAction
+
 from object_database.web.cells.CellsTestMixin import CellsTestMixin
 
 from object_database.web.cells.util import waitForCellsCondition
--- a/object_database/web/cells/cells.py
+++ b/object_database/web/cells/cells.py
@@ -462,6 +462,7 @@ class Cells:
 
         res = {
             'id': cell.identity,
+            'shouldDisplay': cell.shouldDisplay,
             'replacements': replaceDict,
             'component_name': cell.__class__.__name__,  # TODO: TEMP replace when ready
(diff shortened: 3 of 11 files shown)


object_database/web/cells/non_display/key_action.py
```diff
@@ -0,0 +1,41 @@
...
+    keyCmd str: A string of a key combo like `Alt+i`, `Meta+D`
+                or event single keys like `X`.
+    callback func: A function that will be called with the
+                event response dictionary data as the sole arg.
+    wantedInfo list of str: A list of keys on the UI keydown event object
+                whose values should be sent back to the Cell in the response
+                message. Optional; defaults to None.
+    priority int: The priority level of the event response.
+    stopsPropagation bool: Whether or not this KeyAction should fire and then
+                stop other KeyActions with the same keyCmd from firing
```
braxtonmckee, line 48: does this also prevent the key from propagating to other UX such as a code-editor?

object_database/web/cells/non_display/key_action.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class KeyAction(Cell):
+    """
+    Parameters
+    __________
+    keyCmd str: A string of a key combo like `Alt+i`, `Meta+D`
+                or event single keys like `X`.
+    callback func: A function that will be called with the
+                event response dictionary data as the sole arg.
```
braxtonmckee, line 42: can we document the contents of this dictionary? I'm not sure what would be in there or how I'd use it as a python programmer.

object_database/web/cells/non_display/key_action.py
```diff
@@ -0,0 +1,41 @@
...
+Listens key-combo press events from the UI"""
+from ..cells import Cell
+
+
+class KeyAction(Cell):
+    """
+    Parameters
+    __________
+    keyCmd str: A string of a key combo like `Alt+i`, `Meta+D`
+                or event single keys like `X`.
```
braxtonmckee, line 40: Is the format of strings that are used to describe key combinations standardized (i mean is it a standard web thing)? If so, can we include a link. If not can we describe in detail what will work? Also, how does this work mac vs pc? Also, how to encode arrow keys?

object_database/web/cells/non_display/key_action.py
```diff
@@ -0,0 +1,41 @@
...
+    callback func: A function that will be called with the
+                event response dictionary data as the sole arg.
+    wantedInfo list of str: A list of keys on the UI keydown event object
+                whose values should be sent back to the Cell in the response
+                message. Optional; defaults to None.
+    priority int: The priority level of the event response.
+    stopsPropagation bool: Whether or not this KeyAction should fire and then
+                stop other KeyActions with the same keyCmd from firing
+    """
+    def __init__(self, keyCmd, callback, wantedInfo=None, priority=4, stopsPropagation=False):
```
braxtonmckee, line 58: should we assert `1 <= priority <= 4`?

braxtonmckee (changes requested):
This looks good! Could use a little more explicitness on the docs (mostly because I don't know what the browser expects).  Also needs a rebase and squash.

object_database/web/cells/non_display/key_action.py
darth-cheney: No. This is not possible directly, because Ace editor has it's own bizarre way of dealing with key events. What we can do, however, is start and stop our global KeyAction-based listener whenever a given Ace editor gains or loses focus. I'm working on that now and will push and update when it's up and running -- might be tomorrow the way things are going

object_database/web/cells/non_display/key_action.py
darth-cheney: The `<metakey>+<alphanum>` structure is non standard, but the practice seems to be common enough. If you would like I can add this to the comments somewhere in the Cell module.
  
The possible key values -- ie the parts after a `+` or just represented on their own, can be found on MDN here:
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key
  
For arrows the names are `ArrowRight` `ArrowLeft` `ArrowUp` `ArrowDown`
  
Mac vs PC is tough for me to test out -- I've never been able to get this repository to build and install on Mac. Daniel and I tried early on but we were spinning our wheels so left that effort behind

object_database/web/cells/non_display/key_action.py
darth-cheney: I'm not sure this is clear, but the dictionary data is requested when the Python user instantiates the `KeyAction` Cell. By default it will just be an empty dictionary since by default you don't pass any names of data you want from the browser.

object_database/web/cells/non_display/key_action.py
darth-cheney: I think I have just pushed a solution. No need for a new test in Playground: just click into the CodeEditor and the keycombo (Alt+t) will stop working. Blur from it (click away) and it should start working again.

darth-cheney:
@braxtonmckee I have responded to your review here and also pushed some requested changes you should know about:
1. Better documentation about `keyCmd` values and event data you can request, now in the `key_action.py` module as commended notes
2. A blur/focus capability on `CodeEditor` components, that will force our custom action listener to stop when the editor has focus and to start again when an editor loses focus. This should prevent clashes between registered key listeners in Ace vs KeyAction. They are now totally separate in both implementation and effect.

braxtonmckee:
OK, that sounds good. Needs rebase and squash. I think the code looks good. Testing mac/pc, you just need to run the browser on a mac, not the server, if you want to test out how the key combos will come across...

darth-cheney:
@braxtonmckee squashed and pushed. CI seems to be failing from some HTTP error but the "real" checks all seem to be passing

braxtonmckee:
Yeah, that's happening more and more. Some pip/coverage stuff.

Issue #114, opened 2019-06-24
NamedTuple should have a `replacing` method


braxtonmckee:
Create a keyword argument method `replacing` so that `t.replacing(k1=v1, k2=v2, ...)` returns a copy of the tuple with member 'k1' replaced with value 'v1', etc. This makes code that just wants to get a copy of a tuple with a single field updated much cleaner to read.

braxtonmckee:
In order to implement this, you'll need to override `typeMethodsConcrete` in the `PyNamedTupleInstance` class to generate the new functions.

szymonlipinski:
I have a question:

As the function typeMethodsConcrete is static and gets no arguments, and I need to generate a function for each passed member of the NamedTuple, is there a way to get the list of the passed field names to the NamedTuple constructor from the static function?

braxtonmckee:
So right now, typeMethods gets called with a Type* that's the actual type.
We're dropping that when we call typeMethodsConcrete. We should be passing
that through everywhere, so that the type methods can vary with the type.
You should also cast it to 'py_instance_type::modeled_type*'. Then the
concrete function in PyNamedTuple will get a NamedTuple* and you can use
that.

szymonlipinski:
I have made a new commit in a branch https://github.com/APrioriInvestments/nativepython/commit/165a45c2cc384950f0ccf3226918bf362de5c4f5. I'm not really sure how to implement all the parts to fit properly.

braxtonmckee:
I made some comments. Hopefully they give you enough to proceed.

szymonlipinski:
Thank you for your comments. I have added some more changes https://github.com/APrioriInvestments/nativepython/compare/szymon_add_tuple_methods. The main problem now is that the arguments passed to both constructor functions are wrong, so it doesn't compile now.

braxtonmckee:
I made some additional comments directly on the commit in question.

szymonlipinski:
Thank you. I have a couple of question while trying to understand the code:

1. Is the eltPtr a pointer to an element in an internal list?
2. In the Type class, the constructor function gets two arguments: `(int64_t count, const ptr_func& ptrToChild)` and then it loops from 0 to the count, passing to the second argument. How does it work?
[...]

PR #155, opened 2019-07-06
Add named tuple replacing function

szymonlipinski:
Add NamedTuple::replacing function.

Description is in the ticket #114.

## Motivation and Context
The `NamedTuple` will be able to use the function like `t.replacing(k1=v1, k2=v2, ...)`, which returns a copy of the tuple with member 'k1' replaced with value 'v1', etc.

## How Has This Been Tested?
There are new tests for checking the function, including compiler integration, and reference counts.

## Types of changes

## Checklist:

### DIFF
--- a/nativepython/tests/tuple_compilation_test.py
+++ b/nativepython/tests/tuple_compilation_test.py
@@ -107,3 +107,59 @@ class TestTupleCompilation(unittest.TestCase):
             tupToString(Tuple(int, str)((0, 'a'))),
             ["0", "a"]
         )
+
+    def test_named_tuple_replacing_error(self):
+        """We should have errors for all the field names passed to the replacing function,
+        if the fields are not in the tuple definition.
+        """
+        NT = NamedTuple(a=int, b=str)
+        n1 = NT(a=1, b='x')
+
+        @Compiled
+        def f1(x: NT) -> NT:
+            return x.replacing(c=10)
+
+        with self.assertRaisesRegex(Exception, "The arguments list contain names 'c' which are not in the tuple definition."):
+            f1(n1)
+
+        @Compiled
+        def f2(x: NT) -> NT:
+            return x.replacing(c=10, d=10, e=10)
+
+        with self.assertRaisesRegex(Exception, "The arguments list contain names 'c, d, e' which are not in the tuple definition."):
+            f2(n1)
+
+    def test_named_tuple_replacing_function(self):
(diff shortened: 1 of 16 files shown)


nativepython/type_wrappers/tuple_wrapper.py
```diff
@@ -150,9 +151,40 @@ def __init__(self, t):
...
         if ix is None:
             context.pushException(AttributeError, "'%s' object has no attribute '%s'" % (str(self.typeRepresentation), attribute))
             return
 
         return self.refAs(context, instance, ix)
+
+    def convert_method_call(self, context, instance, methodname, args, kwargs):
+        if methodname == 'replacing' and not args:
+            return context.push(self, lambda newInstance: self.initializeReplacing(context, newInstance, instance, kwargs))
+
```
braxtonmckee, line 170: need a call to `super()` here in case a bad method call gets in there.

typed_python/PyCompositeTypeInstance.cpp
```diff
@@ -189,4 +189,87 @@ int PyNamedTupleInstance::tp_setattr_concrete(PyObject* attrName, PyObject* attr
...
     return -1;
-}
\ No newline at end of file
+}
+
+/**
+ * Searches for the index of the element in the container.
+ * Returns the position, if the element doesn't exist, returns -1.
+ */
+int PyNamedTupleInstance::findElementIndex(std::vector<std::string> container, std::string &element) {
```
braxtonmckee: need a `const std::vector<std::string>&` reference here.

nativepython/tests/tuple_compilation_test.py
```diff
@@ -107,3 +107,47 @@ def tupToString(x):
...
+
+        @Compiled
+        def f2(x: NT, a: int, b: str) -> NT:
+            return x.replacing(a=a, b=b)
+
+        n3 = f2(n2, 123, '345')
+        self.assertIsNot(n1, n2)
+        self.assertIsNot(n2, n3)
+        self.assertEqual(n3.a, 123)
+        self.assertEqual(n3.b, '345')
```
braxtonmckee, line 153: need a refcount test (analagous to the ones you have below) so that we can verify that native code also refcounts correctly.

braxtonmckee (changes requested):
This is looking good. Almost there, just a couple more things to fix. Also, please squash down to just one commit. Thanks.

PR #168, opened 2019-07-10
[set] initial impl of Set

ynisakov:
Extend typed_python by adding a `Set` data structure/type function.

## Motivation and Context
The typed_python `Set` type function constructs a c++ "Type" object  that models the type and then wraps it in a `PyTypeObject` and exposes it to python.

## Approach
At a high level, a new constant in `Type.hpp` was created to model the category of `Set`. Since the `Set` model type is  binary compatible with `Dict(T,None)`, the implementation approach was simply curated to reflect the difference. Since the key/value pairs are packed into a hashtable and None is a data structure with 0 bytes, the interface was mostly portable with the exception of unique interface behavior that was added that is unique to operating a traditional python `set()`.

## How Has This Been Tested?
There are relevant tests in `types_test.py` that walkthrough testing basic set functionality such as iteration, common interface calls, refcounting. 

## Types of changes

## Checklist:

### DIFF
--- a/typed_python/PyInstance.cpp
+++ b/typed_python/PyInstance.cpp
@@ -42,6 +42,7 @@
 #include "PyOneOfInstance.hpp"
 #include "PyForwardInstance.hpp"
 #include "PyEmbeddedMessageInstance.hpp"
+#include "PySetInstance.hpp"
 
 Type* PyInstance::type() {
     return extractTypeFrom(((PyObject*)this)->ob_type);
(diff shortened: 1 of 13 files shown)


typed_python/PyInstance.cpp
```diff
@@ -1082,6 +1086,8 @@ PyObject* PyInstance::categoryToPyString(Type::TypeCategory cat) {
     if (cat == Type::TypeCategory::catListOf) { static PyObject* res = PyUnicode_FromString("ListOf"); return res; }
     if (cat == Type::TypeCategory::catNamedTuple) { static PyObject* res = PyUnicode_FromString("NamedTuple"); return res; }
     if (cat == Type::TypeCategory::catTuple) { static PyObject* res = PyUnicode_FromString("Tuple"); return res; }
+    if (cat == Type::TypeCategory::catSet) { static PyObject* res = PyUnicode_FromString("Set"); return res; }
+    if (cat == Type::TypeCategory::catConstDict) { static PyObject* res = PyUnicode_FromString("ConstDict"); return res; }
```
braxtonmckee: this const dict line is a duplicate. maybe a merge error?

typed_python/types_test.py
```diff
@@ -2067,3 +2067,141 @@ def test_named_tuple_replacing_refcount(self):
...
+        native_d = Dict(str, Set(int))()
+        i = Set(int)()
+        native_d["a"] = i
+        self.assertEqual(_types.refcount(i), 2)
+        native_d["b"] = i
+        self.assertEqual(_types.refcount(i), 3)
+        del native_d["a"]
+        self.assertEqual(_types.refcount(i), 2)
+        native_d = None
+        self.assertEqual(_types.refcount(i), 1)
```
braxtonmckee, line 2160: We also need a test for Set(TupleOf(int)) where we check the refcount on the tuples we place inside the set. This test is basically testing whether Dict has the right semantics.

braxtonmckee (changes requested):
This looks pretty good! A couple of minor points: we need some tests on the refcounts of the _items_ within "Set" (see my comment below) during each of the different operations (clear, add, discard). Secondly, please amend the commit to have a message "Refs #169" in the first line of the commit message so we can track it (I just added it). Finally, please check the linting.

After that we can merge. Great job!

typed_python/types_test.py
ynisakov: done

typed_python/PyInstance.cpp
ynisakov: yep, cp dup'd. removed.

braxtonmckee (commented):
nice job

PR #174, opened 2019-07-11
WS and Cells Lifecycle

darth-cheney:
This PR implements new message constructors and a standard message format for most Cells lifecycle events, along with appropriate CellHandler changes on the UI side.

## Motivation and Context
There are several reasons for this update:
1. JSON structures sent across the socket are not standardized and therefore can be tricky to debug sometimes
2. `CellHandler` and the components on the front end were not previously "aware" of when Cell objects are created, updated, or discarded (see section below).
3. The adjustments in the PR better position us to get rid of replacements and implement the dual strategy of HTTP/Socket requests for whole structures, followed by atomic (cell specific) updates.
  
### Lifecycle Awareness ###
Prior to the changes in this PR, every time a Cell is updated, recalculated, and a message with that information went across the wire, we initialized a new Component for it on the front end, called its render function, and then the Component object is lost/garbage collected. This means that the same Cell, during any updates while it is "alive", can trigger the initialization of multiple Component instances that it then cannot continue to communicate with.
  
This has already presented problems when it comes to more event/UX heavy components like the datepicker.
  
In short, we need to keep track of component instances for each cell (when created or updated) and to stop keeping track of them when the cell is discarded. This will allow us to build more nuanced lifecycle actions into both the Cells **and** their corresponding UI Components.

### `Messenger` ###
We introduce a new utility module called `Messenger` whose responsibility is to take some parameters (usually a Cell instance and some additional data) and return a JSON-parsable dictionary that has a standard structure for messages going across the wire.
  
`Messenger` functions are now used in the lifecycle portions of `cells.py`, rather than explicitly formatting them in that code.
  
#### Requirements ####
All outbound socket messages require at minimum these two pieces of information:
1. A `channel` key whose value is a string starting with a `#`. For the moment all messages are sent on `#main`. This is not currently a used feature, but we might need it in the near future.
2. A `type` key whose value is a unique string starting with a `#`. This tells the UI (and devs) what the "action" is that the message corresponds to.
  
#### Message Types ####
For the moment, we have the following message types:
  
| Type | Lifecycle Meaning |
| -------- | -------------------------- |
| `#cellUpdated` | A Cell has either been created or has been recalculated and we need to send over new replacements / children / namedChildren and some updated exportData to the corresponding Component in the UI |
| `#cellDiscarded` | A Cell is no longer in scope and any reference to it should be removed from the UI |
| `#appendPostscript` | Tells the UI's CellHandler to append a single Python-formatted JS script to its internal collection of postscripts, which will be executed later |
  
### CellHandler Changes ###
The `CellHandler` on the front end has been updated to switch on these new message types and execute the appropriate internal handler for each one. Right now those internal methods have the same names as the type, so `#cellUpdated` will end up calling `handler.cellUpdated()` etc.
  
Note that we also have two catch-all handlers:
* `invalidMessageFormat()` -- Called when some JSON structure comes over the socket and it doesn't have a `type` property for us to switch on;
* `doesNotUnderstand()` -- Called when a valid JSON message comes over the wire, but the handler doesn't have an internal method ready to deal with it.
  
#### Keeping Components Alive
Additionally, CellHandler now keeps a dictionary of Component instances mapped by Cell id, so that we know if a component for a Cell already exists in the UI, we simply need to "update" it rather than re-creating a whole new instance. This allows us to also have lifecycle methods on the Components, and will alleviate dataflow problems between Cells and their UI Components.
  

## How Has This Been Tested?
Cells tests and linting have been run. We have also tested this using the demo frontend.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:


object_database/web/content/CellHandler.js
```diff
@@ -110,6 +135,61 @@ class CellHandler {
...
+        }
+
+        // If the cell has a domNode property,
+        // it means it's a hyperscript element
+        // and this is what we should deal with.
+        if(foundCell && foundCell.domNode !== undefined){
+            foundCell = foundCell.domNode;
+        }
+
+        if(foundComponent){
```
dkrasner, line 172: do we need to remove `this.cells[message.id]` as well?

object_database/web/content/CellHandler.js
```diff
@@ -140,113 +220,118 @@ class CellHandler {
...
+            }, message.extraData);
+            let nextPropsToPass = component.componentWillReceiveProps(
+                component.props,
+                nextProps
+            );
+            component._updateReplacements(message.replacementKeys);
+            component._updateProps(nextPropsToPass);
+            //let cClass = this.components[message.componentName];
+            //var component = new cClass(nextProps, message.replacementKeys);
+            var velement = component.render();
```
dkrasner, line 245: do we need to call the projector to re-render the element here?

object_database/web/content/CellHandler.js
```diff
@@ -140,113 +220,118 @@ class CellHandler {
...
+        } else {
+            // In this case, we are creating this cell for the first
```
dkrasner, line 247: this is a `cellUpdated` type of message - why would there be a new cell to create here?

object_database/web/content/CellHandler.js
darth-cheney: We need to keep it around for now. Here's why:
* `this.activeComponents` is a dictionary that stores component instances by CellID
* `this.cells` is a dictionary that stores *rendered* hyperscript or DOM elements by CellID.
  
We currently still need both because the projector needs to know where to replace things.

object_database/web/content/CellHandler.js
darth-cheney: There is no `cellCreated` message from the Python side at the moment. For now, the handler can easily tell whether or not we are updating an existing cell component or creating a new one based on the id's presence or absence from the `this.activeComponents` or even `this.cells` dictionaries.

object_database/web/content/CellHandler.js
darth-cheney: No, because it happens down here:
https://github.com/APrioriInvestments/nativepython/blob/eric-cells-lifecycle/object_database/web/content/CellHandler.js#L285
 
In the particular if/else block you are referring to, we are simply setting the `velement` variable, which gets projected later on in the block I've linked to here.

PR #197, opened 2019-07-19
refactor dict and set tp layout

ynisakov:
DictType and SetType share the same underlying typed python memory layout. The code can be refactored out and centralized to reduce maintenance.

## Motivation and Context
Dict and Set type functions have the same underlying hash table memory layout. The centralized code can be maintained in a singular header to reduce slippage of bugs permeating through the two type functions.

## Approach
This is a refactor that includes a test suite to independently test the hash table layout exclusive of various typed_python constructs. The hash table layout now lives in its own header file. 

## How Has This Been Tested?
New test suite has been added to test the hash table implementation at the cpp layer. Those can be found in `typed_python/direct_types/DirectTypesTest.cpp`. These are basic at best but good enough as a starting point to allow us to further make adjustments to the hash table implementation with some bit of confidence.

## Types of changes

## Checklist:


braxtonmckee (approved):
This looks good. I might, consider leaving 'layout' as a member of the DictType and SetType and just typedef it to the common definition, since that's a little closer to how we're using it, but not a big deal.

PR #198, opened 2019-07-19
Initial Cell Styling Removal / Cleanup

darth-cheney:
The PR is a first pass at moving most styling from the Python side to the frontend side. We have also added new Cell/Components and extra styling adjustments as needed.

## Motivation and Context

Previously, Cells had internal methods to set their own custom inline CSS styling on the resulting DOM elements (things like `width()` and `_divStyle()`). Our goal is to eventually have a simple vocabulary on a cell by cell basis for what visual properties a cell should have, rather than to set all kinds of styling in Python where it cannot be tracked or managed easily.
  
Another goal is to have most styling in CSS file(s) with classnames and cascades that make sense for the front end.

## Approach

We have removed all Cell calls to the `_divStyle()` method. Whatever styles were required for each Cell, we have moved into the Component for that Cell as an inline style (the next phase will involve moving these things to the stylesheet/CSS classes).
  
In the process of doing this we also refactored some Cell/Components and created a couple of new ones.
  
### `Sequence` ###
The `Sequence` Cell is used all over the place. Its expected behavior is to display some collection of Cells stacked inline vertically. It can also use the `+` operator on any Cell to wrap that Cell in a `Sequence`.
  
One issue is that the UIs were using `Sequence`, but the resulting elements were displayed in a horizontal row. This had to do with the custom `divStyle()` etc methods that were being called in those circumstances. They were likely a simple hack to get things to display in a row.
  
`Sequence` will now result in exclusively *vertical* displays. Anything that was previously using Sequence but intended for horizontal display will need to be adjusted.
  
### `HorizontalSequence` ###
We introduce the `HorizontalSequence` Cell type which is more or less polymorphic with `Sequence` except for two key factors:
1. The resulting objects will be displayed in a *horizontal* element (rather than vertical)
2. You can use the `>>` operator on any two Cells to automatically wrap them in a `HorizontalSequence`.
  
We have adjusted to UI in places where we saw that Sequence was being used for horizontal things. There might be others that we have missed or do not know about.
  
### `PageView` ###
We have also created a new view level Cell/Component combination called `PageView` which knows how to make the `ActiveWebService` page views -- ie, a menubar at the top and a body taking up the rest of the space. This is now the default in `ActiveWebService`
  
### `ActiveWebService` Refactoring` ###
We have decomposed some of the more complex declarative Cell structures for pages into smaller functions and Cell combinations (in addition to the new `PageView`).
  
### **Heads Up** ###
Because we do not have access to all UIs currently used in production, these changes will likely result in odd looking results in some cases, most likely where Sequences have been used. These will need to be reported to us and probably shared as a playground example.

## How Has This Been Tested?
We have tested this by examining the test frontend.
  
We have also added many new examples to the Playground to ensure that individual Cells are rendering as expected.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

### DIFF
--- a/object_database/web/ActiveWebService.py
+++ b/object_database/web/ActiveWebService.py
@@ -30,7 +30,8 @@ from object_database import ServiceBase, service_schema
 from object_database.web.AuthPlugin import AuthPluginBase, LdapAuthPlugin
 from object_database.web.LoginPlugin import LoginIpPlugin
 from object_database.web.ActiveWebService_util import (
-    Configuration, LoginPlugin, mainBar,
+    Configuration, LoginPlugin,
+    makeMainView,
     displayAndHeadersForPathAndQueryArgs,
     writeJsonMessage,
     readThread
@@ -281,8 +282,8 @@ class ActiveWebService(ServiceBase):
 
     def displayForPathAndQueryArgs(self, path, queryArgs):
         display, toggles = displayAndHeadersForPathAndQueryArgs(path, queryArgs)
-        return mainBar(display, toggles, current_user.username,
-                       self.authorized_groups_text)
+        return makeMainView(display, toggles, current_user.username,
+                            self.authorized_groups_text)
 
     @login_required
     def mainSocket(self, ws, path):
--- a/object_database/web/ActiveWebService_util.py
+++ b/object_database/web/ActiveWebService_util.py
@@ -20,7 +20,7 @@ from object_database.web.ActiveWebServiceSchema import active_webservice_schema
 from object_database.web.cells import (
     Main, Subscribed, Sequence, Traceback, Span, Button, Octicon,
     Padding, Tabs, Table, Clickable, Dropdown, Popover, HeaderBar,
-    LargePendingDownloadDisplay
+    LargePendingDownloadDisplay, PageView
 )
 
 from object_database.web.AuthPlugin import AuthPluginBase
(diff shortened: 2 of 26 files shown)


dkrasner:
@braxtonmckee this has been rebased and commits squashed

braxtonmckee:
So, going through it, I noticed a few things:

First, at the main service page level, the 'search' octicons are on the next line:

image

They're encoded with 'nowrap' which I know we are getting rid of. This chain doesn't have the >> operator implemented yet. So maybe we don't need to fix this yet?

Second, 'press alt t' doesn't work for me.

Third, the 'toggle modal' button fills the screen. If you remove the hidden modal, it doesn't fill the screen. So something about the css is off there.

Fourth, in the 'some buttons' page, the buttons fill the space. It's fine for them to stack vertically (we need the >> operator to make them layout horizontally) but they shouldn't fill like that (and they don't if there's only one).

Fifth, horizontal and vertical split sequences both lay out the same for me.

Sixth, the vertical resize panel can't be resized because it's the smallest it can be. The way this screen ought to work is that the whole screen is divided top-and-bottom with the bottom third being the area to select which cells demo and edit the code. The cell should get the top two thirds. A vertical resize panel should then consume all the vertical space it can.

I tried our main app, but I can't get very far because the buttons fill the space and make the thing impossible to use. So maybe let's fix that and then I'll see what else I can fix.

@darth-cheney - you said you had the >> operator working somewhere else? I see this branch has the horizontal and vertical Sequence objects, so adding the sugar should be fairly easy.

I rebased and squashed this change on top of the main.bundle.js removal change (Which is now in).

dkrasner:
@braxtonmckee did you rebase and squash before or after testing?

braxtonmckee:
Before

braxtonmckee:
I just tried this on the version before I force pushed and I'm seeing the same problems.

darth-cheney:
@braxtonmckee On my local machine I have a copy of this branch from right when I originally pushed for this PR (I have not pulled changes). It seems to be working fine on my end. I have branched from this version (to save it, and the commit history) and pushed it as `eric-divstyle-remove-saved`. It should be identical to what was first PR'd.
  
Here's what I see when I run it:
Screenshot from 2019-07-31 13-41-58
  
Screenshot from 2019-07-31 13-43-00

  
I can confirm, however, that a fresh install with the version that was force-pushed this morning has all kinds of styling problems. I'm thinking it must have been the rebase that took place.

darth-cheney:
I'm doing a fresh rebase on a different branch and this PR is now referenced as #206

Issue #167, opened 2019-07-10
Implement convert-to-python for all types in the compiler


braxtonmckee:
Right now we can convert int/float/bool to `object` because they have simple types that we can invoke directly from the python C api. But we need to be able to convert all typed-python types to `object`, as well as inline constants like `type` or `int`. This will require being able to get a Type* in native code for a particular typed_python type, and to lookup builtin python types by name. 

This isn't totally trivial because we want the compiled shared-object to be loadable by other python instances, so we can't simply hardcode the Type* as an inline pointer value.  Instead, we should use `getOrSetTypeResolver` to set a type-resolver for the current codebase when we load the shared-object library and then ensure that each type in the compiled code has a meaningful name we can use to look up its runtime type. We will also need to implement some kind of static storage in our llvm-code-gen so that we only hit the interpreter once for each type.

Once we have this, wrapping any given typed_python value a python wrapper should be pretty straightforward.

After we do this, it should be fairly easy to make sure that we can compile pretty much anything, because we'll have a coherent framework for taking an operation we can't compile and deferring to the interpreter.

braxtonmckee:
This is merged.

PR #200, opened 2019-07-25
Conversions to and from python objects.  Refs #167 #202

rubidge:
## Motivation and Content
Improve scope of compilation with ability to convert typed-python types to object.
Allow bytes literals.

## Approach
This uses the approach of passing a raw type pointer (Type*) to identify the type.  (Not the getOrSetTypeResolver as mentioned in #167.)

## How Has This Been Tested?
Back and forth conversions produce same value.
Bytes literals are used in compiled functions.

## Types of changes

## Checklist:

### DIFF
--- a/nativepython/type_wrappers/python_object_of_type_wrapper.py
+++ b/nativepython/type_wrappers/python_object_of_type_wrapper.py
@@ -13,11 +13,13 @@
 #   limitations under the License.
 
 from nativepython.type_wrappers.wrapper import Wrapper
-from typed_python import Int64, UInt64
+from typed_python import Int64, Int32, Int16, Int8, UInt64, UInt32, UInt16, UInt8, Float32, Float64
 from nativepython.typed_expression import TypedExpression
 
 import nativepython.native_ast as native_ast
+from nativepython.native_ast import VoidPtr
 import nativepython.type_wrappers.runtime_functions as runtime_functions
+from typed_python._types import getTypePointer
 
 
 class PythonObjectOfTypeWrapper(Wrapper):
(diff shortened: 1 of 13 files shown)


typed_python/_runtime.cpp
```diff
@@ -299,20 +300,139 @@ extern "C" {
         return floorresult;
     }
 
-    PyObject* nativepython_runtime_int_to_pyobj(int64_t i) {
+    // returns layout*
+    void *np_runtime_pyobj_to_typed(PyObject *obj, void* tgt, uint64_t tp) {
```
braxtonmckee: tp should be a Type*, and tgt an instance_ptr. It will do the same thing (llvm will pass everything to us in 64 bit registers) but it makes the code clearer.

nativepython/type_wrappers/python_object_of_type_wrapper.py
```diff
@@ -91,18 +93,36 @@ def convert_to_type_with_target(self, context, e, targetVal, explicit):
...
             context.pushEffect(
                 targetVal.expr.store(
-                    runtime_functions.pyobj_to_int.call(e.nonref_expr)
+                    arith_calls[t].call(e.nonref_expr)
                 )
             )
             return context.constant(True)
 
-        if target_type.typeRepresentation == UInt64:
+        tp = getTypePointer(t)
```
braxtonmckee: please make this a method on ExpressionConversionContext, so that we will end up having exactly one place where this function gets called.

braxtonmckee (changes requested):
Please make a couple of minor changes. Otherwise, looks good.

typed_python/_runtime.cpp
rubidge: Regarding tgt: tgt is a pointer to a layout, and &tgt is an instance_ptr. So I wasn't sure what type to assign to tgt to represent pointer to a layout.

rubidge: These changes have been made, this branch is superseded by jeff-conv4, pull request #208.

rubidge:
Superseded by pull request #208

Issue #191, opened 2019-07-16
Implement string joining in the compiler


braxtonmckee:
We have string 'split' but no corresponding 'join' method.

This should be implemented as a 'convert_method_call' method. The actual string joining should be done in C++. Like the implementation of `split` which defers to `runtime_functions.string_find` which calls `nativepython_runtime_string_split` in `_runtime.cpp`.

In this case, you'll produce a function like 

```
void nativepython_runtime_string_join(StringType::layout* outString, StringType::layout* separator, ListOfType::layout* toJoin)
```

that defers to functions implemented in StringType to do the actual work. You should be able to look at the split code for examples. 

We mimic python string internals: each string uses 1,2, or 4 bytes per codepoint depending on the largest codepoint in the string. You'll need to compute the number of bytes per codepoint across all the strings, compute the appropriate size of the output string, and then bitcopy all the relevant substrings in (making sure to cast each string to the max number of bits used as you go).

The string_wrapper needs to make sure it passes a `ListOf(str)`, but the function should work for any iterable. So it should first pass the argument into a function (using `call_py_function`) that checks whether its argument is `ListOf(str)` and if not, use `call_py_function` to call something that iterates the arguments, verifies they're strings, and pushes them into a ListOf(str).

szymonlipinski:
The layout class is.
```
    class layout {
    public:
        std::atomic<int64_t> refcount;
        typed_python_hash_type hash_cache;
        int32_t count;
        int32_t reserved;
        uint8_t* data;
    };```

The passed as list is `ListOfType::layout* toJoin`.

How can I iterate through the list if I don't have the size of the element?

**Comment by braxtonmckee** [comment]
In this case, you know you are joining a ListOf(str), so the element Type*
is String::Make(). You must ensure that this is the case from the calling
context.  Also, for the most part, the best way to interact with the layout
object is to produce the relevant list type and use that.  For instance,
you can see StringType::split has code like

```
    static ListOfType* listofstring = ListOfType::Make(StringType::Make());
    listofstring->resize((instance_ptr)&outList, 0, 0);
```

which you can see is producing the type for ListOf(str) `listofstring` and
using it to interact with the layout.

szymonlipinski:
Thanks. It looks like there is something wrong with the code, currently it complains that the returned NoneType cannot be converted to String. https://github.com/APrioriInvestments/nativepython/compare/szymon-dev

PR #204, opened 2019-07-29
Add string::join. Ref: #191

szymonlipinski:
## Motivation and Context

Adds string.join method to the compiler. It supports calling as:

```
separator.join(items)
```

where:

* separator is a string
* items is either ListOf(str) or TupleOf(str)

## How Has This Been Tested?

There are new tests.

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/nativepython/tests/string_compilation_test.py
+++ b/nativepython/tests/string_compilation_test.py
@@ -12,7 +12,7 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
-from typed_python import Function, ListOf, _types
+from typed_python import Function, ListOf, _types, TupleOf, Dict, ConstDict
 from typed_python.test_util import currentMemUsageMb
 from nativepython.runtime import Runtime
 import unittest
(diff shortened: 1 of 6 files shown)


nativepython/tests/string_compilation_test.py
```diff
@@ -457,3 +457,67 @@ def test_find_perf(f, s, subs):
...
+        # test longer separator
+        res = f("---", ListOf(str)(["1", "2", "3"]))
+        self.assertEqual("1---2---3", res)
+
+        # test longer items
+        res = f("---", ListOf(str)(["aaa", "bb", "c"]))
+        self.assertEqual("aaa---bb---c", res)
+
+        # test empty separator
+        res = f("", ListOf(str)(["1", "2", "3"]))
```
braxtonmckee: need also to test some empty strings in the list itself

nativepython/tests/string_compilation_test.py
```diff
@@ -457,3 +457,67 @@ def test_find_perf(f, s, subs):
...
+        with self.assertRaisesRegex(TypeError, ""):
+            f_tup_int(",", ListOf(int)([1, 2, 3]))
+
+        # test passing list of other types than strings
+        @Compiled
+        def f_int(sep: str, items: ListOf(int)) -> str:
+            return sep.join(items)
+
+        with self.assertRaisesRegex(TypeError, ""):
+            f_int(",", ListOf(int)([1, 2, 3]))
```
braxtonmckee, line 561: we should check that this also works on a Dict(str, str) or a ConstDict(str, str), since both of those are currently iterable, and we need to verify that theywork.

typed_python/StringType.cpp
```diff
@@ -814,3 +814,96 @@ void StringType::assign(instance_ptr self, instance_ptr other) {
...
+
+    StringType::Make()->constructor((instance_ptr)outString);
+
+    // return empty string when there is nothing to join
+    if(toJoin->count == 0)
+    {
+        return;
+    }
+
+    static ListOfType *inputList = ListOfType::Make(StringType::Make());
```
braxtonmckee: this variable should be called `listOfStrType`. OTherwise it looks like a data variable.

nativepython/tests/string_compilation_test.py
```diff
@@ -457,3 +457,67 @@ def test_find_perf(f, s, subs):
                 .format(c_elapsed, elapsed)
         )
         """
+
+    def check_string_join(self, f):
+        """This function contains tests for implementation of the string join
+        for both: ListOf(str) and TupleOf(str).
+        The tests for the types are as separate functions, so it will be easier
+        to find the failing line later.
+        """
```
braxtonmckee: need some tests with strings that are not just 1 byte per codepoint. Try joining with two-byte and four-byte codepoints in both separator and individual string values.

braxtonmckee (changes requested):
looks good. I requested a couple of variable renames and some more tests.

nativepython/tests/string_compilation_test.py
szymonlipinski: done

typed_python/StringType.cpp
szymonlipinski: done

nativepython/tests/string_compilation_test.py
szymonlipinski: done

szymonlipinski:
One interesting thing is that the ConstDict has different ordering of the items than the Dict. It looks like Dict stores them in the order passed in constructor, while the ConstDict removed the empty string and ordered the keys.

braxtonmckee:
ConstDict does keep things sorted. But what do you mean it removed the
empty string? That sounds like a bug. Is it a problem with the join or the
ConstDict itself?

nativepython/tests/string_compilation_test.py
szymonlipinski: done

szymonlipinski:
I think all is fine with the ConstDict.

PR #205, opened 2019-08-01
[set] add support for difference, copy, union, intersection, minor bug fixes

ynisakov:
The PR enriches the `Set(T)` API with additional commonplace functionality that is expected when working with sets; namely the API was extended to include `difference(), union(), intersection(), and copy()` functions. The functions, where args are appropriate, are not limited to working with only other sets, but can be used with various other iterable container objects from both python and typed_python to increase the general application of the container.

## Motivation and Context
The `Set(T)` type function needs additional api enrichments to mimic the behavior one would expect from using a native python set(). Having an ability to express a difference, intersection, or union, between elements is, at a fundamental level, what makes a set useful. 

## Approach
1 - `Set(T)` now supports `difference(args...), union(args...), intersection(args...),  copy()`, where `args...` is any combination of potentially the following objects `set, list, tuple, ListOf(T), TupleOf(T), Set(T)` 
2 -`Set(T)` also now supports construction from unicode strings e.g. , it is possible to express the following `Set(str)("helloworld")` which will initialize a set of unique characters.

## How Has This Been Tested?
There are a number of additional tests added to `typed_python/types_test.py` to cover the extended API functionality.

## Types of changes

## Checklist:

### DIFF
--- a/typed_python/PySetInstance.cpp
+++ b/typed_python/PySetInstance.cpp
@@ -17,17 +17,22 @@
 #include "PySetInstance.hpp"
 
 PyMethodDef* PySetInstance::typeMethodsConcrete(Type* t) {
-    return new PyMethodDef[6]{{"add", (PyCFunction)PySetInstance::setAdd, METH_VARARGS, NULL},
+    return new PyMethodDef[9]{{"add", (PyCFunction)PySetInstance::setAdd, METH_VARARGS, NULL},
                               {"discard", (PyCFunction)PySetInstance::setDiscard, METH_VARARGS,
                                NULL},
                               {"remove", (PyCFunction)PySetInstance::setRemove, METH_VARARGS, NULL},
                               {"clear", (PyCFunction)PySetInstance::setClear, METH_VARARGS, NULL},
+                              {"copy", (PyCFunction)PySetInstance::setCopy, METH_VARARGS, NULL},
+                              {"union", (PyCFunction)PySetInstance::setUnion, METH_VARARGS, NULL},
+                              {"intersection", (PyCFunction)PySetInstance::setIntersection,
+                               METH_VARARGS, NULL},
+                              {"difference", (PyCFunction)PySetInstance::setDifference,
+                               METH_VARARGS, NULL},
                               {NULL, NULL}};
 }
 
-PyObject* PySetInstance::try_remove(PyObject* o, PyObject* args, bool assertKeyError) {
+PyObject* PySetInstance::try_remove(PyObject* o, PyObject* item, bool assertKeyError) {
     PySetInstance* self_w = (PySetInstance*)o;
-    PyObjectHolder item(PyTuple_GetItem(args, 0));
     Type* item_type = extractTypeFrom(item->ob_type);
     Type* self_type = extractTypeFrom(o->ob_type);
 
@@ -71,7 +76,8 @@ PyObject* PySetInstance::setRemove(PyObject* o, PyObject* args) {
         return NULL;
     }
 
-    return try_remove(o, args, true);
+    PyObjectHolder item(PyTuple_GetItem(args, 0));
+    return try_remove(o, item, true);
 }
 
 PyObject* PySetInstance::setDiscard(PyObject* o, PyObject* args) {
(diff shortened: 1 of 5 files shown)


typed_python/PySetInstance.cpp
```diff
@@ -93,6 +100,288 @@ PyObject* PySetInstance::setClear(PyObject* o, PyObject* args) {
...
+void PySetInstance::copy_elements(PyObject* dst, PyObject* src) {
+    PySetInstance* dst_w = (PySetInstance*)dst;
+
+    PyObjectStealer iterator(PyObject_GetIter(src));
+    if (!iterator) {
+        throw PythonExceptionSet();
+    }
+
+    PyObject* item;
+    while ((item = PyObjectStealer(PyIter_Next(iterator)))) {
```
braxtonmckee: passing this into a temporary PyObjectStealer like this is not a good idea.  I'm not sure what it will do. The 'stealer' pattern is intended to acknowledge that it's constructed with an incref, and needs to decref, and to use RAII to prevent exceptions from creating leaked increfs. But in this case, the PyObjectStealer is temporary, so it's just pre-decreffing the container item before storing it in 'item' which will be fine as long as the item is held by the container, but won't work if this were a generator, or something else that returns an object with only one refcount.

A safer pattern would be to use the 'iterate' function defined in util.hpp. E.g.
```
        iterate(src, [&](PyObject* item) {
            ..body of your loop here...
        });
```

We also need to check whether 'src' is a typed_python Set, Tuple, or ListOf with the same element type, and if so, bypass the python interpreter and insert the elements directly. See, for instance, `PyListOfInstance::listExtend` which checks if we are extending a list with another list of the same type, and if so, uses a much faster path.

typed_python/PySetInstance.cpp
```diff
@@ -93,6 +100,288 @@ PyObject* PySetInstance::setClear(PyObject* o, PyObject* args) {
...
+
+    PyObjectStealer iterator(PyObject_GetIter(other));
+    if (!iterator) {
+        decref((PyObject*)new_inst);
+        return NULL;
+    }
+
+    PySetInstance* self_w = (PySetInstance*)o;
+    PyObject* item;
+    while ((item = PyObjectStealer(PyIter_Next(iterator))) != NULL) {
```
braxtonmckee: same comment as above, on both stealers and on needing a fastpath for other typed-python objects of appropriate type.

typed_python/PySetInstance.cpp
```diff
@@ -213,6 +503,8 @@ void PySetInstance::copyConstructFromPythonInstanceConcrete(SetType* setType, in
                 if (!item) {
                     if (PyErr_Occurred()) {
                         throw PythonExceptionSet();
+                    } else {
```
braxtonmckee: also need to fix the stealer pattern here and use 'iterate'.

typed_python/PySetInstance.cpp
```diff
@@ -236,11 +528,33 @@ void PySetInstance::copyConstructFromPythonInstanceConcrete(SetType* setType, in
                 setType->insertKey(tgt, key.data());
             }
             return;
+        } else if (PyUnicode_Check(pyRepresentation)) {
+            if (PyUnicode_READY(pyRepresentation) < 0) {
```
braxtonmckee: why wouldn't it work to just include PyUnicodeCheck in the checks in line 495? Now that I think about it, shouldn't this work with anything that's iterable?

braxtonmckee (changes requested):
great functionality. I made some specific comments that need to be addressed...

typed_python/PySetInstance.cpp
ynisakov: @braxtonmckee , thanks for feedback. took care of replacing the Stealer pattern with iterate  (simple but sweet function). Also, added fast path branching for typed python types where applicable.

typed_python/PySetInstance.cpp
ynisakov: done

typed_python/PySetInstance.cpp
ynisakov: done

typed_python/PySetInstance.cpp
ynisakov: @braxtonmckee , good point, rolled up PyUnicode processing to the other iterables. There was some reason why I originally spun it out, but once I rebased recently, it seems to work as part of the other iterable code path. Always feels good to clip some code and simplify.

ynisakov:
@braxtonmckee , seems to be a build error, but seems unrelated to my changes.

PR #206, opened 2019-08-01
Divstyle Fixes Rebase (was #198)

darth-cheney:
This PR is a fixed-rebase version of #198


darth-cheney:
@braxtonmckee I have done a detailed rebase here from the "clean" branch I had previously saved and everything seems to be working up to date. The only bug that was persisting was the `KeyAction` one you mentioned and that indeed was a bug that was introduced. I have also fixed that here.
  
Note that I haven't yet squashed the commits but will do so when you feel like the PR is in working order.

braxtonmckee:
This looks much better. One thing I noticed is that the 'split' argument to Sequence doesn't appear to work, and we have demos for it (that don't work), but the HorizontalSequence _does_ work, which is good. So maybe we should just get rid of that argument and use the HorizontalSequence when we want it.

The rest of this looks pretty good. One problem I'm having is that I cannot control the code editor size any more. It looks like 'width' doesn't get honored any more, which is fine, but I need a way of doing the layout. I think this is analagous to the problem of how the resizable Vertical split doesn't work because it doesn't take up the whole page.  

In general, we need all these components that subdivide space to use as much as they can from the parent.  Any component, like the code editor (and also the handsontables) should also fill whatever space is allotted to it. We could make a "Sized" cell that lets us give an explicit pixel height or width (or both) around another cell to control cases where we we want that, although we can probably do most of what we need with the splitters now.

I'll see if I can generate playground examples on this that represent what I'd like...

dkrasner:
When `HorizontalSequence` was added to this PR (see the original PR #198  conversation) all the `split` Sequence arguments should have been removed. This is now done. 


The taking space full space issue is something we are working on in general. The main problem here was the presence of un-styled divs coming from helper cells such as `Subscribed`. For example, if you are a child of a styled element but are wrapped in a `Subscribed`, telling you to take 100% space means taking 100% space of `Subscribed` not the conceptual parent so what you get is harder to predict. Specific examples might be fixable in css, but in general this should be fixed by created accessor "no display" cells which don't add any additional elements to the DOM. `KeyAction` is an example of such a cell.  Perhaps this is not the example problem you are seeing but it is out there.

braxtonmckee:
ha I was just writing you. Yes, this is the central problem. I see the same thing - the unstyled divs end up compressing themselves to be as small as possible and it makes things like the Vertical split not work. Similarly for the text editor.

In general, we have two modes for cells: space filling, and not space filling. (Maybe we could split that into horizontally and vertically and track it separately).  So one way to solve the problem would be to track this on the python side. In this idea, the root is 'space-filling' in both direction, meaning that if you were to put a single card inside of it, that card would fill the space in all directions.  Any time you use '+' to create a Sequence, you should have produced a set of cells that are vertically stacked, and do not fill space (vertically). And the analagous behavior for the horizontal split.  

Because cells know their parents when they are rendered, they can figure out from the surrounding context that they are supposed to fill space horizontally but not vertically.  If you use SplitView, then the intent is to fill space. If you create an explicitly sized cell (say, we made a SizedContainer cell or something), then cells are intended to fill space internally again.  "Text" and "Button" and other leaf cells would never fill space.

I guess my point is that the way we build ux with these things, there are basically two modes: from the root, we're taking the visual space and chopping it up with boxes.  At some point, we get down to 'leaf' mode, where we have some text or buttons, or whatever, and we want to stack them, and have the nodes just flow.  So we just need a coherent set of expectations on this and we should be good.

But ultimately, I do need to be able to say something like

```
def someBody():
    return cells.ResizeablePanel([
        textEditor(),
        someDisplay()
    ])

cells.SplitView([
    (someHeaderBar(), 1),
    (someBody(), 8),
    split='horizontal'
])
```

and end up with a display where the outer cells.SplitView takes up all the space, and the text editor and plot / data display take up all the space inside of their components.  Right now, I can't seem to do that and get the text editor or plots to take up the right amount of space.

What do you think?

dkrasner:
That makes sense. 

I think the first thing to do is to get rid of all "non display" cells rendering as DOM elements. Then see what issues are left; it will then be clear how to solve them, server, client etc. 

I am not sure that should be part of this PR as it already introduces a bunch of changes and it would be good to see how they play out in the real application. But if you can't really move forward with the real application because we are breaking things here then just let me know and we'll work here, rebasing as needed on dev.

braxtonmckee:
I do think we need to be able to control this before I can take the PR,
because the nasty manual hacks I had in before where I explicitly set
height and width don't work any more.

I think removing the intermediate divs could be pretty complicated. If you
have a subscribed in another subscribed, and the inner subscribed changes,
it seems pretty tricky to me to figure out how to represent that update.
But I agree that if you can do this it would be good...

OTOH The convention I described shouldn't be too hard to implement - you
just look at your parent to figure out if you're space filling in H and V,
and if so, adjust your height and width to be 100% accordingly.  You can
probably do all of this on the client side too, and it should be a less
intrusive change.

darth-cheney:
@braxtonmckee
> I do think we need to be able to control this before I can take the PR, because the nasty manual hacks I had in before where I explicitly set height and width don't work any more.
  
Can you provide an example (maybe a screenshot?) of the TextEditor problem you are having on this branch?
  
The solution for me was to simply use SplitView and give the portion that held the editor whatever proportion of the total width I wanted. That works well and is dynamic. As you said, HorizontalSequence is not space filling but SplitView is. Unless I am missing something there should already be enough capability to compose the layouts as you would need them, no? 
  
Of course if that's the expectation and it's not working, then there must be a bug somewhere and I can take a look. I just need to be pointed to an example if possible.
  
> I think removing the intermediate divs could be pretty complicated. If you have a subscribed in another subscribed, and the inner subscribed changes, it seems pretty tricky to me to figure out how to represent that update. But I agree that if you can do this it would be good...
  
Yes this will be pretty much impossible in the current rendering setup. However, once we merge in some of the current work on refactoring the socket and whole rendering cycle (I'm working on this on another branch and it's coming along), we will have a lot more control and can do this with moderate fuss.

braxtonmckee:
OK, maybe we are a little closer than I thought. I just pushed up a change that reworks the CellsTestService display to not use explicit sizing. You have a SplitView that divides results from selection, and then an internal split that divides between the selector and the text editor. I see that it's taking the full size which is good. 

I think what we need for me to be able to take this change is 
* I would expect that the first SplitView (results vs 'inputArea') would fill vertical space. The fact that it's not is the reason that the BasicVResize panel doesn't work.
* By default we should not be allowing overflow / scrollbars to just pop up anywhere. But in this UI, for example, if you have too many demos on the page, you will need a scrollbar in the input space. This should be "cells.Scrollable" which I see exists, but I'm not sure if it works since we don't have any examples of it, and we just removed the use of `_divStyle` (which we should get rid of if we don't use it)

Once we have that working (which should be pretty easy) I should be able to replicate the other examples from the main UI that are not working in the playground.

Does that sound reasonable?

dkrasner:
I spent a while on this and there are still issues with making sure that children of specific elements fill space appropriately. Some of this has to do with `Subscribed` divs floating around, some of this has to do with browser computed styles. I don't really see what is forcing these height computes, and there is no information of source in the devtools, but I suspect it has to do with flex. For example, our flex proportion properties are not working right, i.e. you should be able to say how much of the parent div a child should take relative to the other children but I am not seeing any response here. 

I'll continue hacking at this. 

adding the `Scrollable` element should be good. I think it should be `display: none` element that is simply there to insist that the wrapped element has `overflow: auto` but I'll think about that.

dkrasner:
@braxtonmckee ok take a look now

a few things to note. There were a number of changes which might affect the production app (in ways I can't see from the web service test). 

* there was doubling up on `main` elements which I have removed
* `container-fluid` (being a grid based bootstrap class) was clashing and preventing some of the flex grow behavior we want, so now this wrapper is removed from the `Main` component. 
  * Note: it is still around in some of the others: `SubscribedSequence`, `CollapsiblePanel`, `Tabs`, `Columns`  - perhaps these should also be removed or adjusted, but I am not seeing much in the way of major issues at the moment
* `Subscribed` elements/cells now are told to take 100% height and width of the parent, so then their children can adjust accordingly - this seems to have solved a number of issues at least in the short-term
* general layouts have been adjusted in the web service test 
* You will notice that BasicVResize doesn't look like BasicHResize, but this is just an illusion because `Card` is not taking up full height; what you are seeing is true vertical resize on the row elements there. I didn't want to remove `Card` b/c it's the one cell/component we have with nice borders and without then it's hard to see where on the display to click to begin adjusting 
   * `Card` is a bootstrap class, maybe we want to override it to take full height, or maybe get rid of the bootstrap version of the class replacing it with our own - unsure
* Since we want everything to grow, the elements will continually adjust in size depending on their content. This by definition prevents scrolling. For the moment I added a `height` argument to the `Scrollable` cell so that you can stop the element from growing and scroll inside it. You can see an example in `cells_demo`. I am not sure this is what we want in the long run. Maybe we are going to want special "no-grow" containers, but that will force the python developer to be more aware of layout specifics. For example, showing a "no-grow" container could break some of the flexible down the DOM tree etc. 

anyhow play around with this, see what else breaks let me know 


a few screenshots of the current:
Screen Shot 2019-08-08 at 2 29 30 PM
Screen Shot 2019-08-08 at 2 29 04 PM
Screen Shot 2019-08-08 at 2 28 53 PM

braxtonmckee:
OK, cool I'll play with this. I should be able to get the main app to work reasonably with this format now.

braxtonmckee:
OK, this is getting closer. Some issues I'm seeing:

On the card not filling space, I think that's OK. Cards should take up H space, but not V space.  We can make a separate thing called 'Panel' which takes up both.

On the scrollables: I'm not sure what the intended behavior is.  What I had been thinking is that a Scrollable doesn't have a 'height' - it simply fills the space it's given (in terms of its own size) and then allows the wrapped cell to take up as much space as it wants within the interior scroll area.

Right now, it looks like it jams everything down into a tiny space (the card edges are compressed). I see something like this:
image

in nonscrollable, if I make 40 cards, the whole UI gets a scroll bar at the outside level (which we don't want). I think the desired behavior here should be that the cells get as much space as they'd like, but if they float off the edge, they just don't display at all.

Second, the MultiPageTable isn't obeying the rules. It's much too big, and then somehow we end up with a scrollbar in both directions at the outermost scope of the page, which shouldn't happen once we're inside of a SplitView.

It looks like we lost some of the padding between the Cards?

The semantics of >> with buttons look a little weird. Buttons, text, etc, are all supposed to take up the minimal space required for their internals. Somehow the combo with the sequences is breaking that because if you have buttons a, b, c, and you write (a+b+c) you get something meaningful, but (a >> b)  causes the buttons to fill vertical space and (a+b) >> c causes the first two buttons to display correctly, and the third button to become right-aligned and fill vertical space. My intent with + and >> is for all objects that are involved in such an expression to use the minimal amount of space possible and to be upper-left justified, almost as if they were laid out as text.

The collapsible panel should be filling space as much as possible. Right now it looks like (in the demo at least) it's getting placed at the bottom of its allocation?

Cards in the horizontal split sequence should not take up all vertical space.

dkrasner:
I am not seeing exactly what you are saying (webpack rebuild? browser hard refresh?):

here are the two examples of scrollable that I see:  

Screen Shot 2019-08-09 at 4 55 34 PM
Screen Shot 2019-08-09 at 4 55 24 PM


There is definitely padding on those Card elements (set to `2` in the example screenshots). Maybe you mean margin? there isn't any margin on the Card, but I don't remember that being an argument/option in the past, unless "padding" was supposed to mean "margin" .... regardless we haven't changed that since the `Card.js` component was created. 


In general for scrolling you need to have a bounding element, and we are flex growing everything to fill space appropriate. Maybe we need a cell that that explicitly bounding (which seems like Scrolling could very well be), maybe we need to see if it's possible to set `page_root` to 100% viewport height and make sure that nothing ever grows out of it, but that would require a more careful layout planning/use on the part of the python dev, i.e. you could have everything overflow over everything else and render a total mess. Whatever you think here?

MultiPageTable fills the entire view space b/c there is no styling on it, i.e. it's raw space where whatever cell you code in the editor will display as is, i..e if you want to wrap it in something with padding like a `Card` you can and it will have boundary. It's not scrolling inside the table b/c there is nothing telling the Table not to get bigger than X so it defaults to scrolling the entire page (this is true in general). 

I don't see some of the other issues you are referring to. I looked through all of the test service, and did find some issues but not the ones you mention. Maybe you can add an example or change one of the examples with a screenshot of what you are seeing?

braxtonmckee:
I rebuilt and hard-refreshed, and somehow am still having these issues. I'm
in chrome.  Is it possible that the CSS is wrong or not rebuilt or
something? I do see a log message 'PropertyValidator.js:9 Scrollable >>
height has an empty description!' in the console, not sure if that's
relevant.

braxtonmckee:
Also, I'm running 'make build-js' to rebuild the webpack. Possible that's not working?

dkrasner:
so you are not seeing what i am seeing, screenshot wise - that's very weird, am in chrome as well (exclusively for all this testing) 

i supposed it's possible webpack didn't rebuild; manually remove the bundle, it's under `web/content/dist/main.bundle.js` and then build again - either it will not load at all, since there is no bundle or it will load the new one. At least that will confirm that you are seeing the latest on this branch

the log message 'PropertyValidator.js:9 Scrollable is coming b/c one of the scrolling examples fails to pass in the argument - currently those property validators are warnings and don't affect anything we are talking about here

dkrasner:
@darth-cheney can you try rebuilding on this branch - do you see the screenshot braxton posted above for the scrolling card examples in `CellsTest` - what i see are the screenshots i posted

we are both on chrome

darth-cheney:
@braxtonmckee @dkrasner 

I have just pulled down and done a fresh install here, and this is what I see:
Screenshot from 2019-08-09 15-09-34
  
Screenshot from 2019-08-09 15-12-22
  
Note that I have tried this both with the "primary" dev install method (`make install` and virtualenv) and using the make recipes we made in the recent PR for Braxton. Both seem to show the same thing on my end.
  
@braxtonmckee To answer your question about bundle size: for the moment we are not minifying code (stripping whitespace, auto-shortening function/variable names) and we are likely adding "sources", which allows for better debugging. Additionally, we have several third party libraries bundled in with our code.
  
Because of the way bundling works, we never really look at the bundle file directly (and therefore there isn't much use in sharing it to track down the problem).
  
Can you paste the terminal output for your bundle build command?
  
If it isn't the bundle then it must be either a hard refresh on the browser is needed (Shift + Cmd/Ctrl + R) or there is some weird environmental issue that we haven't thought of.

braxtonmckee:
```
braxton@braxton-Wild-Dog-Pro:~/code/nativepython$ make build-js
. .nodeenv/bin/activate; \
cd object_database/web/content; \
npm run build

> content@1.0.0 build
/home/braxton/code/nativepython/object_database/web/content
> webpack

Hash: e33b17756a86210ef7a7
Version: webpack 4.35.0
Time: 607ms
Built at: 08/09/2019 4:15:17 PM
         Asset      Size  Chunks             Chunk Names
main.bundle.js  2.14 MiB    main  [emitted]  main
Entrypoint main = main.bundle.js
[./CellHandler.js] 14.9 KiB {main} [built]
[./CellSocket.js] 8.86 KiB {main} [built]
[./ComponentRegistry.js] 3.29 KiB {main} [built]
[./main.js] 1.94 KiB {main} [built]
[./node_modules/moment/locale sync recursive ^\.\/.*$]
./node_modules/moment/locale sync ^\.\/.*$ 3 KiB {main} [optional] [built]
[./node_modules/webpack/buildin/module.js] (webpack)/buildin/module.js 497
bytes {main} [built]
    + 180 hidden modules
```

darth-cheney:
@braxtonmckee That output looks correct, and it appears to be generating a valid bundle without issue.
  
Can you try these steps:
1. Re-clone `eric-divstyle-rebase` in a new directory somewhere on your system and navigate to the root dir of the repo;
2. Run `make install`
3. Run `source .venv/bin/activate`
4. Run `python ./object_database/frontends/object_database_webtest.py`
  
This will also build a new bundle as a part of the install process into a fresh virtual environment. Then we can know for sure whether or not this is some strange environmental issue.

braxtonmckee:
just tried that, still looks the same on my box.  and I can verify that I
can change the javascript and see changes in the resulting components. so I
know the build is working.

dkrasner:
what's your version of chrome?

braxtonmckee:
I'm on 67.0.3396.99

dkrasner:
i am on 76.0.3809.100 (Official Build) beta (64-bit) but i can't imagine that there are such drastic difference. 

can you check out what you are seeing on http://167.71.84.31:8000  - it's a small digital ocean droplet i spun up - i did nothing other than run `make install` and then built the bundle by running `webpack` in `/object_database/web/content`  and what i am seeing there matches what i am seeing locally

braxtonmckee:
My home machine is on chrome 69. It also has the problem (messed up cards in the scrollable.). Firefox seems to do the right thing. Let me try upgrading. Although if that turns out to be the difference, we should try to figure out what's wrong, because version 69 is not so old..

braxtonmckee:
OK, upgrading solves the problem.  That seems very suspicious to me though. We must be hitting some pretty weird edge case, because everything in that layout is pretty vanilla, no?

dkrasner:
Also i just realized that i didn't address you scrollbar question: the scrollbar is **on** the `Scrollable` DOM element, not outside of it. The reason that you were seeing the scrollbar all the way to the right, as if it was at the pageview /root level, was because `Scrollable` was taking up 100% width (all elements take up 100% width by default). I added another example `multiScrollable` which makes this clearer (you can see it live [here](http://167.71.84.31:8000/services/CellsTestService?category=scrollable&name=multiScrollable)). 

In general if we want things to scroll at certain prescribed moments we need to bound the height. Bounding the view height to the viewport is not going to be enough as page elements will all try to flex and grow likely causing surprising behavior. So we want something that will bound elements vertically, and possibly horizontally as well. Maybe this should be a special element, maybe this should be part of scrollable as it already is.. 

what do you think?  can jump on a call to discuss.

dkrasner:
ah just seeing your upgrade comment - yes very weird, we are just using flex nothing special from what i can tell and it should all be supported (https://caniuse.com/#feat=flexbox) 

if it were one of the js libs we used for resizing that would be more suspect but here i really can't see what it would be

braxtonmckee:
On the scrollable, my intent is that you can write

```
cells.SplitView([
    (Scrollable(someCell), 1),
    (someOtherCell, 1),
    split='vertical'
])
```

in which case, the main space is split 50/50 (top and bottom panel). Bottom panel does not have a scroll bar. `someOtherCell` is expected to either further subdivide the space, or, if it involves sequences and the items can't be fit, to get clipped.  `Scrollable(someCell)` is expected to render without scroll bars, unless `someCell` gets too big, in which case we should inject a scrollbar within the space allocated to the scrollabe. Clearly this is possible because it's how text editors work. I really don't want to have to have explicit height requirements in there.  I'm not sure how it ought to be implemented - maybe you need a dom element inside the scrollable that's inset a little that's the actual scrollable, so that if the scrollbar appears suddenly it doesn't create a layout change outside the scrollable.

dkrasner:
" Scrollable(someCell) is expected to render without scroll bars, unless someCell gets too big,"

so this is how overflow works in general (the property we use to scroll); no scrollbars will render until an element gets too big, and "too big" means bigger than its allotted space. 

```
In order for overflow to have an effect, the block-level container must have either a set height (height or max-height) or white-space set to nowrap.
```
from https://developer.mozilla.org/en-US/docs/Web/CSS/overflow


all text editors i've seen have height or max-height set somewhere along the way (ace falls into this category)

another way to approach this is to have a number of set page layouts, i.e. ones where the basic design template is there, div/cell/containers are set with their appropriate and the python dev fills in the blanks - within the template we we set appropriate max height properties

braxtonmckee:
check out the email I just sent you.  We should be able to get a formal
height from the structure of the cells.  Exactly how to expose that to CSS
I don't know.

dkrasner:
rebased and squashed version [here](https://github.com/APrioriInvestments/nativepython/pull/217)

Issue #202, opened 2019-07-25
Compile bytes literals in expressions


rubidge:
This involves adding an ast.matches.Bytes case to convert_expression_ast in expression_conversion_context.py.  Looks like pythonObjectRepresentation() can already handle bytes, so should be easy.

braxtonmckee:
This is merged.

Issue #185, opened 2019-07-16
typed_python `Class` instances should use `__repr__` when available


braxtonmckee:
right now we just use a generated __repr__ and don't check whether the user has supplied one.

rubidge:
Would this be an appropriate test case?
```
        @Compiled
        class ClassWithRepr(Class):
            def __repr__(self):
                return "repr"

        self.assertEqual(repr(ClassWithRepr()), "repr")
```

braxtonmckee:
I don't think the @compiled thing will work on an entire class, although
that's something we should implement. We need to test and implement both
_using_ the repr in normal typed python, and also compiling it. So
something like

```
class ClassWithRepr(Class):
    def __repr__(self):
        return "repr"
    def __str__(self):
        return "str"

self.assertEqual(str(ListOf(ClassWithRepr)([ClassWithRepr()])), "[str]")
self.assertEqual(repr(ListOf(ClassWithRepr)([ClassWithRepr()])), "[repr]")

@SpecializedEntrypoint
def callRepr(x):
    return repr(x)

self.assertEqual(callRepr(ClassWithRepr()), "repr")
```

etc.

braxtonmckee:
This is closed.

PR #208, opened 2019-08-08
Conversions to and from python objects. Refs #167 #202 #185

rubidge:
## Motivation and Context
Improve scope of compilation with ability to convert typed-python types to object.
Allow bytes literals.
Allow repr() and str() on typed python Classes.

## Approach
This uses the approach of passing a raw type pointer (Type*) to identify the type. (Not the getOrSetTypeResolver as mentioned in #167.)

## How Has This Been Tested?
Back and forth conversions produce same value.
Bytes literals are used in compiled functions.
repr() and str() work on typed python Classes with custom __repr__ or __str__.

## Types of changes

## Checklist:

### DIFF
--- a/typed_python/_runtime.cpp
+++ b/typed_python/_runtime.cpp
@@ -340,18 +340,14 @@ extern "C" {
         return floorresult;
     }
 
-    // returns layout*
-    uint8_t *np_runtime_pyobj_to_typed(PyObject *obj, uint8_t* tgt, Type* tp) {
-        // tgt is a layout pointer
+    void np_runtime_pyobj_to_typed(PyObject *obj, instance_ptr tgt, Type* tp) {
         PyEnsureGilAcquired acquireTheGil;
-        PyInstance::copyConstructFromPythonInstance((Type *)tp, (instance_ptr)&tgt, obj, true);
-        // returns modified layout pointer
-        return tgt;
+        PyInstance::copyConstructFromPythonInstance(tp, tgt, obj, true);
     }
 
-    PyObject* np_runtime_to_pyobj(PyObject *obj, uint64_t tp) {
+    PyObject* np_runtime_to_pyobj(instance_ptr obj, Type* tp) {
         PyEnsureGilAcquired acquireTheGil;
-        return PyInstance::extractPythonObject((instance_ptr)&obj, (Type*)tp);
+        return PyInstance::extractPythonObject(obj, tp);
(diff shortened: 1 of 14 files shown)


typed_python/_runtime.cpp
```diff
@@ -303,20 +340,139 @@ extern "C" {
         return floorresult;
     }
 
-    PyObject* nativepython_runtime_int_to_pyobj(int64_t i) {
+    // returns layout*
+    uint8_t *np_runtime_pyobj_to_typed(PyObject *obj, uint8_t* tgt, Type* tp) {
+        // tgt is a layout pointer
+        PyEnsureGilAcquired acquireTheGil;
+        PyInstance::copyConstructFromPythonInstance((Type *)tp, (instance_ptr)&tgt, obj, true);
```
braxtonmckee: you need to be passing tgt, not `&tgt`.  Imagine Type is NamedTuple(x=int,y=int). This takes up 16 bytes. As it currently stands, you will take the NamedTuple out of the python object and jam it into the 16 bytes of stackspace that have been allotted for `tgt`, which is 8 bytes. 

This function ought to read

```
void np_runtime_pyobj_to_typed(PyObject* ob, instance_ptr tgt, Type* tp) {
    PyEnsureGilAcquired acquireTheGil;
    PyInstance::copyConstructFromPythonInstance(tp, tgt, obj, true);
}
```

Then, when you call it, just drop the 'store' operation.  Take the NamedTuple case. You're already passing it 'targetInstance.expr' , which is in fact a NamedTuple*. You cast it to void* to pass it to this function, and it comes in as 'tgt'.

braxtonmckee (changes requested):

rubidge:
Thanks, will make this change...

typed_python/_runtime.cpp
```diff
@@ -303,20 +340,135 @@ extern "C" {
         return floorresult;
     }
 
-    PyObject* nativepython_runtime_int_to_pyobj(int64_t i) {
+    void np_runtime_pyobj_to_typed(PyObject *obj, instance_ptr tgt, Type* tp) {
+        PyEnsureGilAcquired acquireTheGil;
+        PyInstance::copyConstructFromPythonInstance(tp, tgt, obj, true);
```
braxtonmckee, line 345: what happens if this conversion fails. Do we have tests for this case?

PR #211, opened 2019-08-09
Fixups in preparation for switching from nose to pytest

atzannes:
## Motivation and Context
I had found a strange behavior in nose where it would sometimes (but deterministically) fail to run some tests when the testsuite was split into chunks for parallelization. I traced the problem to a bug in `nose`. The bug was not simple to fix and I noticed that nose has gotten no updates in the last 5 years, so it seems quite deprecated. `pytest` seems to be today's golden standard and it seems capable to run tests written for `nose`.

## Approach
I installed `pytest` locally and ran `pytest --collect-only` which runs the test discovery step, and it came back with a small number of warnings. This PR makes the necessary changes to fix those warnings.

## How Has This Been Tested?
TravisCI is happy and `pytest --collect-only` gives no warnings (or errors)

This is ready to merge as soon as the tests pass


atzannes:
Ready for review @braxtonmckee

PR #213, opened 2019-08-13
Switch to using pytest instead of nose on TravisCI

atzannes:
## Motivation and Context
- `nose` is deprecated and buggy; `pytest` is the status-quo. For example, we have noticed that `nose` can forget to run some of the tests, or can forget to run the `setUpClass` method for some tests causing them to fail.

## Approach
`pytest` can run testsuites written for `nose`, so the transition was straightforward.

## How Has This Been Tested?
- All the tests pass with pytest
- `unittest.skip` was honored by pytest but the tests didn't show up as skipped (with nose the tests didn't show up at all). By adding the decorator`@pytest.mark.skip(reason="xyz")` both testing frameworks do the right thing: `test.py` (aka `nose`) still skips these tests silently, and `pytest` shows that the tests exist and that they were skipped during testing.
- One of the skipped tests (in database_test.py), seems to pass both with pytest and with nose, so it was re-enabled
- After all these, it looks like pytest is running one test more than `./test.py`, but it's not clear which one that is
- Wrote a test (see code below) to check that stdout and stderr are captured correctly. `pytest` worked correctly:
image
- `nose` didn't do as well (without the -v flag we don't see the output, and with it the output is mangled along with nose's output):
image

- Here's the test:
```
    def test_sys_write(self):
        import sys
        sys.stdout.write("Printed to STDOUT")
        sys.stderr.write("Printed to STDERR")
        self.assertFalse(True)

```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,9 +29,11 @@ jobs:
         - pipenv lock --dev --requirements > dev-reqs.txt
         - pip install -r dev-reqs.txt
         - make lint
+
     - name: "pipenv check"
       script:
         - pipenv check || (sleep 60; pipenv check) || (sleep 60; pipenv check)
+
     - name: "Unit Tests"
       before_script:
         - ulimit -c unlimited -S       # enable core dumps
@@ -41,7 +43,7 @@ jobs:
         - sudo apt-get install -y gdb  # install gdb
         - make testcert.cert
         - coverage erase
-        - ./test.py -s
+        - pytest
         - ls -la
         - coverage combine
       after_success:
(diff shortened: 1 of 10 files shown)


braxtonmckee:
So, you'd recommend just ditching test.py altogether?

atzannes:
Not sure what `test.py` offers that `pytest` doesn't @braxtonmckee, you're probably more knowledgeable. For example, I see a `--dump_native` flag. Generally speaking, if we can use exclusively `pytest` that will probably be a good thing.

braxtonmckee:
yes, we have some flags but we can probably control those using environment
variables, (or maybe there's something in pytest that can be used to
communicate configuration to tests during a run?)

atzannes:
I think environment variables would be a good way to do what some of our flags do, e.g., `--dump_native`.
Beyond that, `pytest` allows you to parameterize tests and fixtures, but I don't know much about that:
- https://docs.pytest.org/en/latest/parametrize.html
- http://doc.pytest.org/en/latest/example/parametrize.html

PR #216, opened 2019-08-22
Black code formatter as a pre-commit hook

atzannes:
## Motivation and Context
Good code hygiene is very valuable, but doing it consistently across a team is challenging and frustrating. Black code formatter does a good job and almost fully automates that task. 

`pre-commit` allows determining which scripts will execute on devellopers' machines prior to committing to git and allows managing and updating them using a repo-config file. The only requirement is that each dev runs `pre-commit install` once for their local repo.

This PR sets up `pre-commit` and prepares most of the configuration for black code formatter, but the black format script is left commented out until we apply black to the entire codebase.

## Approach
Relying on individual developers to remember to run the Black code formatter on their modified code before each commit is not very realistic, but luckily black provides a hook that works with `pre-commit`, a convenient utility for installing and managing git hooks.

The preferred way to install the hooks is by running `make install`. That will install `pre-commit` in the virtual environment and invoke `pre-commit install`. This step need to happen whenever the configuration file of pre-commit changes (`.pre-commit-config.yaml`).

## First time use
The first time the pre-commit hook executes the black code formatter on a file you changed, you may want to separate the (many) changes the formatter does from your edits to make reviewing the code possible. You can achieve this by running:
- `git commit`: this will trigger the black formatter which will modify your edited files and fail to commit
- now you have your set of changes staged and black's changes not staged
- commit your changes using `git commit --no-verify`
- stage black's changes and commit them. This time the black commit-hook should run and succeed.

## How Has This Been Tested?
This has been tested locally using the procedure described above (see First time use)

## Checklist:


atzannes:
@braxtonmckee thoughts?

atzannes:
@braxtonmckee I pinned the version of `black` that gets checked out to the current release.

atzannes:
got :+1:  from Brax by email

Issue #212, opened 2019-08-12
Ensure that Alternative implements all the magic methods correctly


braxtonmckee:
Right now you can write

```
A = Alternative("A", X={}, __str__=lambda self: "a string")

str(A.X())
```

and get the correct dispatch. Similarly with __add__. But everything else is not implemented.

rubidge:
Which of the remaining magic methods should be supported?
__new__, __init__, __del__
__eq__, __ne__, __lt__, __gt__, __le__, __ge__, __hash__
__bytes__, __format__
__getattr__, __getattribute__, __setattr__, __delattr__, __dir__
__get__, __set__, __delete__, __set_name__, __slots__
__instancecheck__, __subclasscheck__, __class_getitem__
__length_hint__
__getitem__, __setitem__, __delitem__
__missing__, __iter__, __reversed__
__i*__ (in-place operators) 
__r*__ (reversed operators)
__complex__, __int__, __index__,__float__
__round__, __trunc__, __floor__, __ceil__
__enter__, __exit__, __await__
__aiter__, __anext__, __aenter__, __aexit__

braxtonmckee:
I think we can exclude anything related to class initialization or async, or that's not implemented for classes. So, of your list,

these are definitely in:

    eq, ne, lt, gt, le, ge, hash bytes, format, getattr, getattribute, setattr, delattr, dir

these are just for decorators. we might implement them for classes, but I'm not sure it makes sense for alternatives:

    get, set, delete, set_name, 

not these:

    slots, instancecheck, subclasscheck, class_getitem

we don't generate this:

    length_hint

this is just for subclasses of dict:
  
    delitem missing,

these all seem like we would do it

    getitem, setitem,  iter, reversed, i* (in-place operators), r* (reversed operators), complex, int, index,float,round, trunc, floor, ceil, enter, exit

we're not going to deal with async at this point. maybe at some point in the future:

    await aiter, anext, aenter, aexit

braxtonmckee:
This is merged.

Issue #210, opened 2019-08-08
Make sure all of our builtin types convert to 'bool' correctly


braxtonmckee:
In compiled code, we should be able to write

```
def f(d: Dict(int, int)):
    if d:
        ...
```

and similarly for all of `Tuple`, `ConstDict`, `TupleOf`, `ListOf`, `Class` (it hits `__bool__`), &etc

braxtonmckee:
This is merged.

Issue #209, opened 2019-08-08
Objects with type 'object' should be convertible to 'bool'


braxtonmckee:
e.g. this should work:

```
def f(x: object):
  if x:
      ...
```

this should call 'PyObject_IsTrue' under the hood (after reacquiring the gil)

braxtonmckee:
This is closed.

PR #218, opened 2019-08-23
Provide conversions to bool for many types. Refs #210, #209, #212.

rubidge:
## Motivation and Context
* Provide bool conversion functionality.

## Approach
    * including Classes with __bool__ or __len__ methods
    * including Alternatives with __bool__ or __len__ methods
* Also fix some string conversions:
    * Avoid buffer overflow in nativepython_int64_to_string.
    * Provide suitable uint64 conversion to string.
* and some cosmetic changes

## How Has This Been Tested?
* Compare direct bool() with compiled bool and SpecializedEntrypoint
* Note there is a failing case (outer_specialized_bool) in python_obect_of_type_compilation_test.py that warrants further investigation.

## Types of changes

## Checklist:

### DIFF
--- a/nativepython/tests/python_object_of_type_compilation_test.py
+++ b/nativepython/tests/python_object_of_type_compilation_test.py
@@ -13,8 +13,8 @@
 #   limitations under the License.
 
 from typed_python import (
-    Function, TupleOf, ListOf, String, Dict, NamedTuple, OneOf,
-    Bool,
+    Class, Member, Function, Tuple, TupleOf, ListOf, String, Bytes, ConstDict, Dict, NamedTuple, Set,
+    Alternative, OneOf, NoneType, Bool,
     Int8, Int16, Int32, Int64,
     UInt8, UInt16, UInt32, UInt64,
     Float32, Float64,
@@ -24,6 +24,7 @@ import unittest
 import psutil
 from nativepython.runtime import Runtime
 from typed_python._types import refcount
+from nativepython import SpecializedEntrypoint
 
 
 def Compiled(f):
@@ -34,6 +35,16 @@ def Compiled(f):
 class HoldsAnA:
     def __init__(self, a):
         self.a = a
+
+
+class AClassWithBool(Class):
+    x = Member(int)
+
+    def __init__(self, i):
+        self.x = i
+
+    def __bool__(self):
+        return self.x != 0
 
 
 class TestPythonObjectOfTypeCompilation(unittest.TestCase):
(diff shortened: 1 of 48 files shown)


nativepython/tests/python_object_of_type_compilation_test.py
```diff
@@ -134,3 +145,180 @@ def fro_and_to(x):
...
+            def inner_specialized_bool(x) -> bool:
+                return bool(x)
+
+            r1 = bool(x)
+            r2 = compiled_bool(x)
+            r3 = inner_specialized_bool(x)
+            # TODO: outer_specialized_bool will error unpredictably on AClassWithBool if there are many specializations prior to it
+            # it is fine if it is listed among the first 20 or so entries in test_cases
+            # but it fails (error or segfault) if it is near the end of the list
+            # r4 = outer_specialized_bool(x)
```
braxtonmckee, line 230: this is important. please file a ticket for this, ideally with a minimal repro test in the issue

nativepython/type_wrappers/arithmetic_wrapper.py
```diff
@@ -396,6 +402,32 @@ def convert_to_type_with_target(self, context, e, targetVal, explicit):
             )
             return context.constant(True)
 
+        # elif target_type.typeRepresentation == String:
```
braxtonmckee, line 405: let's remove this.

typed_python/_runtime.cpp
```diff
@@ -479,34 +491,70 @@ extern "C" {
...
         ReprAccumulator acc(s);
-        acc << i;
+        acc << f;
 
         std::string rep = s.str();
         return StringType::createFromUtf8(&rep[0], rep.size());
     }
 
-    StringType::layout* nativepython_float32_to_string(float i) {
+//        ReprAccumulator acc(s);
```
braxtonmckee, line 530: can we drop this comment?

nativepython/type_wrappers/alternative_wrapper.py
```diff
@@ -71,6 +72,27 @@ def convert_copy_initialize(self, context, target, toStore):
...
+        target_type = targetVal.expr_type
+
+        if target_type.typeRepresentation == Bool:
+            tp = context.getTypePointer(e.expr_type.typeRepresentation)
+            if tp:
+                if not e.isReference:
+                    e = context.push(e.expr_type, lambda x: x.convert_copy_initialize(e))
+                context.pushEffect(
+                    targetVal.expr.store(
+                        runtime_functions.instance_to_bool.call(e.expr.cast(VoidPtr), tp)
```
braxtonmckee, line 89: why are we hitting a C function here? This will be very slow in a tight loop.  If the alternative has this method, we should be able to dispatch to it directly from compiled code. Similarly for Class objects.

braxtonmckee (changes requested):
looks great - a couple of comments need to get dropped, and we need to dispatch to __bool__ from compiled code, not go through an interpreter call.

nativepython/type_wrappers/alternative_wrapper.py
rubidge: Ok, thanks, I'm looking into how to use call_py_function here.

nativepython/type_wrappers/arithmetic_wrapper.py
rubidge: Yes, I should have cleaned this up earlier.

nativepython/tests/python_object_of_type_compilation_test.py
rubidge: https://github.com/APrioriInvestments/nativepython/issues/219

rubidge:
Superseded by #225.

PR #221, opened 2019-09-02
Some PyPi packaging scripts

atzannes:
## Motivation and Context
We want to get ready to upload a `nativepython` package to pypi

## Approach
Figure out how to upload packages to `test.pypi` and try to install the package from there and use its functionality. For simplicity, we only provide source packages, which means that the user will have to compile them when `pip install`ing them. Otherwise we will have to provide different wheels for different target architectures.

* `make sdist` will create a distribution according to the version in `setup.py`.
* `make testpypi-upload` uploads the distributions under `/dist` created by the previous step to test.pypi. 

## How Has This Been Tested?
* https://test.pypi.org/project/nativepython/
* In a fresh docker container running ubuntu:18.04, installed python3, python3-pip, python3-virtualenv, created a virtual environment, successfully installed nativepython and manually tested in a REPL that I can import OneOf, create OneOf types, and create objects of those types.

The command used for installing the nativepython package from test.pypi was: 
```pip install --extra-index-url https://test.pypi.org/simple/  nativepython==0.0.1dev2```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,8 @@ object_database/web/content/dist
 /.coverage*
 /.tox
 /.env
+/pip-wheel-metadata
+/dist/
 
 # don't need the package lock for this sub-pipenv.
 object_database/web/content/package-lock.json
--- a/INSTALLATION.md
+++ b/INSTALLATION.md
@@ -1,6 +1,19 @@
 ## Development ##
 There are several methods for building and setting up a development environment.
-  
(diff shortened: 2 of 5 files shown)


atzannes:
@braxtonmckee this is ready to merge. I made some documentation reorganization as well in the last commit.

braxtonmckee:
This is superceded.

PR #222, opened 2019-09-02
WIP: Run TravisCI tests on Python 3.6 & 3.7

atzannes:
## Motivation and Context
Because we rely on internal python implementation choices which could change between minor python versions, we should check against minor versions to be more confident our code works with them.

## Approach
Add a job in the `test` stage that is a replica of the job that runs the tests on Python 3.6, but to run tests on Python 3.7

## How Has This Been Tested?
On TravisCI. It looks like some tests fail on Python 3.7 (Hence the WIP status of the PR)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: xenial
+dist: bitonic  # ubuntu:18.04 LTS
 
 language: python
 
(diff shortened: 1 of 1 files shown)


braxtonmckee:
This is superceded by recent work.

Issue #219, opened 2019-08-28
SpecializedEntrypoint with a Class argument causes an error


rubidge:
SpecializedEntrypoint with a Class argument causes an error during compilation.   

Errors are
```
TypeError: 'AClass' object is not iterable"
SystemError: <function specialized_fail> returned a result with an error set
```
which occur at the "return f(*args)" line of inner() within SpecializedEntrypoint() in runtime.py.

```
Test code
    def test_boolconv_specialization_fail(self):

        class AClass(Class):
            pass

        test_cases = [
            TupleOf(int)(),
            AClass(),
        ]

        @SpecializedEntrypoint
        def specialized_fail(x):
            return False

        for x in test_cases:
            r = specialized_fail(x)
```

rubidge:
Funny thing
```
        test_cases = [
            AClass(),
            TupleOf(int)(),
            AClass(),
        ]
```

will succeed.

rubidge:
test_cases = [
            ListOf(int)(),
            AClass(),
        ]

also fails, but succeeds if you replace ListOf(int)() with 

Tuple(int,int)(1,2)
Dict(int,int)()
ConstDict(int,int)()
Set(int)() or
a Concrete Alternative instance

rubidge:
Doesn't need to be a Class; fails with this too:
```
        class AClass():
            pass
```

rubidge:
Does not appear to be a recently introduced change in behavior.
Tested dev branch at 006294 (8/22/2019), fde064 (7/15/2019), and 58bf2c (7/4/2019), and error is the same.

rubidge:
Generated code
```
*************
def py.py.specialized_fail.dispatch(return=*void,input=**void): #->void
    (cast(return,*uint1))[0]=call(py.specialized_fail->uint1)(cast(((input).gep(0s32)).load,**TupleOfLayout))
*************


*************
def py.specialized_fail(x=**TupleOfLayout): #->uint1
    void
    return 0u1
    void
*************


*************
def py.py.specialized_fail.1.dispatch(return=*void,input=**void): #->void
    (cast(return,*uint1))[0]=call(py.specialized_fail.1->uint1)(cast(((input).gep(0s32)).load,**void))
*************


*************
def py.specialized_fail.1(x=**void): #->uint1
    void
    return 0u1
    void
*************
```

rubidge:
[...]

PR #225, opened 2019-09-11
Provide conversions to bool for many types. Refs #210, #209, #212, #219.

rubidge:
## Motivation and Context
* Provide bool conversion functionality.

## Approach
    - including Classes with bool or len methods
    - including Alternatives with bool or len methods
* Dispatch len(), repr(), in, unary operations, binary operations, comparison operations, abs(), `__call__`,  `__hash__`, `__round__`, `__trunc__`, `__floor__`, and `__ceil__` appropriately for both compiled and interpreted Alternative types.
* Magic methods for Alternative Types that can be interpreted (but not compiled):` __bytes__, __format__, __getattr__, __setattr__, __delattr__, __getattribute__, __dir__, __getitem__, __setitem__, __iter__, __next__, __reversed__, __complex__, __index__.`
* When compiling, implement default Alternative comparison operators to match interpreted behavior. 
* Also fix some string conversions.
* Avoid buffer overflow in nativepython_int64_to_string.
* Provide suitable uint64 conversion to string.
* Avoid SpecializedEntrypoint compilation error when noniterable type occurs after an iterable type.
* Fix some refcount leaks involving extractPythonObject.
* and some cosmetic changes

## How Has This Been Tested?
* Compare direct bool() with compiled and SpecializedEntrypoint versions.

## Types of changes

## Checklist:

### DIFF
--- a/nativepython/type_wrappers/alternative_wrapper.py
+++ b/nativepython/type_wrappers/alternative_wrapper.py
@@ -14,7 +14,7 @@
 
 from nativepython.type_wrappers.wrapper import Wrapper
 from nativepython.type_wrappers.refcounted_wrapper import RefcountedWrapper
-from nativepython.type_wrappers.arithmetic_wrapper import FloatWrapper
+from nativepython.type_wrappers.arithmetic_wrapper import FloatWrapper, IntWrapper
 import nativepython.type_wrappers.runtime_functions as runtime_functions
 
 from typed_python import NoneType, _types, OneOf, Bool, Int32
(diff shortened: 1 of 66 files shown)


nativepython/type_wrappers/alternative_wrapper.py
```diff
@@ -71,6 +72,42 @@ def convert_copy_initialize(self, context, target, toStore):
...
+
+        if target_type.typeRepresentation == Bool:
+            alt = e.expr_type.typeRepresentation
+            if hasattr(alt.__bool__, "__typed_python_category__") and alt.__bool__.__typed_python_category__ == 'Function':
+                assert len(alt.__bool__.overloads) == 1
+                y = context.call_py_function(alt.__bool__.overloads[0].functionObj, (e,), {})
+                if y is None:
+                    return context.constant(False)
+
+                ret_type = y.expr_type.typeRepresentation
```
braxtonmckee: it might return 'object' here, which we then need to try to convert to bool. So I think you could get away with

```
return y.expr_type.convert_to_type_with_target(context, y, targetVal, False)
```

rubidge: Yes, that works!

rubidge:
Superseded by PR#234.

PR #234, opened 2019-10-03
Support Alternative magic methods, Bool conversions. Refs #209, #210, #212, #219.

rubidge:
## Motivation and Context
Convert builtin types and objects to Bool.
Support all relevant magic methods for Alternatives.
Along the way, remove some gaps in magic method support for other types, e.g. format(), dir(), both interpreted and compiled.

## Approach
Specifics:
        -* including Classes with` __bool__` or `__len__` methods
        -* including Alternatives with `__bool__` or `__len__` methods
    * For interpreted Alternative types, dispatch the following appropriately:
        -* `__contains__, __len__, __repr__, __hash__, __bytes__, __format__,`
          `__getattr__, __setattr__, __delattr__, __getattribute__, __dir__,`
          `__dir__, __getitem__, __setitem__, __iter__, __next__, __iter__, __reversed__,`
          `__complex__, __int__, __index__, __float__, __round__, __trunc__, __floor__, __ceil__,`
          `__enter__, __exit__`
    * For Alternative types, compile the following magic methods:
        -* `__contains__, __len__, __hash__, binary operators, inplace binary operators, unary operators, __pow__`
          `__call__, __abs__, comparison operators,`
          `__round__, __trunc__, __floor__, __ceil__,`
          `__format__, __bytes__,`
          `__getattr__, __setattr__, __delattr__, __dir__`
        -* Not compilable yet: `__complex__, __index__, __getattribute__, __enter__, __exit__, reverse binary operators`
    * For Register types, add interpreted and compiled support for round, trunc, floor, and ceil.
    * For Bytes type, support bytes().
    * For Dict type, correct parameter order for dict_remove_key call.
    * For all types, compile default dir() matching interpreted dir.
    * For all types, compile default format() matching interpreted format.
    * Avoid SpecializedEntrypoint compilation error when noniterable type occurs after an iterable type.
    * Take more care with cast/convert distinction
    * Fix some refcount leaks that occurred after calling extractPythonObject.
    * Fix compiled` __repr__, __str__, __contains__` pyobject leaks.
    * Fix some string representations:
        -* Avoid buffer overflow in nativepython_int64_to_string.
        -* Provide suitable uint64 string representation.
        -* Correct float64 string representation to more closely match python str(float).
            --* There still are discrepancies.
    * Improved Floatxx -> Intxx conversion
    * and some cosmetic changes
        -* Cleanup: replace` __div__ `with `__truediv__`.

## How Has This Been Tested?
Added tests for bool conversion, for magic methods for Alternatives, and for other builtins for other types (e.g. format(), dir())

## Types of changes

## Checklist:

### DIFF
--- a/nativepython/type_wrappers/alternative_wrapper.py
+++ b/nativepython/type_wrappers/alternative_wrapper.py
@@ -56,22 +56,13 @@ class SimpleAlternativeWrapper(Wrapper):
         return self.layoutType
 
     def convert_hash(self, context, expr):
-        y = self.convert_call_method(context, "__hash__", (expr,))
+        y = self.generate_method_call(context, "__hash__", (expr,))
         if y is not None:
             return y
         tp = context.getTypePointer(expr.expr_type.typeRepresentation)
         if tp:
             return context.pushPod(Int32, runtime_functions.hash_alternative.call(expr.nonref_expr.cast(VoidPtr), tp))
         return None
-
-    def convert_destroy(self, context, target):
-        pass
-
-    def convert_assign(self, context, target, toStore):
-        assert target.isReference
-        context.pushEffect(
-            target.expr.store(toStore.nonref_expr)
-        )
 
     def convert_copy_initialize(self, context, target, toStore):
         assert target.isReference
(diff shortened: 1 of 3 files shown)


braxtonmckee (commented):
This looks great! One thing we will need to do a second pass on is the distinction between cast and convert, particularly for bool / float.  The semantics I implemented in the new function dispatch code are that 'cast' is what we do when we say something like `str(x)` or `float(x)`, and convert is what we do when we try to match `f(x: str)` or `f(x: float)`.  In some cases, it looks like we would allow an Alternative to match a function that accepts a float, and I don't think we want that to happen without an explicit user cast. But I think we should do that in a separate commit (with a test showing the bug). The rest of this looks great.

rubidge:
Thanks, I was wondering about that.  I wrote the test case and modified the
AlternativeWrapper to move the float case from
convert_to_type_with_target() to convert_cast().

But the conversion still happens implicitly.  E.g. I defined an alternative
A with __float__ and a function f(x: float).  When I call
Compiled(f)(A.a()), I want a TypeError but I get f(__float__(A.a()))
instead.   Is this the right test case?

The conversion is happening in PyRegisterTypeInstance.  Comments in that
code assumes isExplicit=true to indicate a cast (might be my own comment),
and this code is used to pass other tests.  So I'll have to think about how
to make changes...
-Jeff

Issue #254, opened 2019-10-15
Compiled Dict.get produces TypeError


rubidge:
The following code fails with the error
{TypeError}Can't call Dict(Int64->Int64).get with args of type (Int64, Int64)

```
        @Entrypoint
        def dict_get(x, y):
            return x.get(y, -1)

        x = Dict(int, int)()

        dict_get(x, 1)
```
(The interpreted code functions properly.)

braxtonmckee:
@rubidge is this fixed now?

rubidge:
Yes -- I can find the test and close this.

rubidge:
The test is test_dict_get_default.

Issue #240, opened 2019-10-09
Ensure that Class implements all the magic methods correctly


rubidge:
If a Class defines a magic method, ensure it is dispatched and handled appropriately, for both interpreted and compiled code.

PR #255, opened 2019-10-21
Support magic methods for Class types.  Refs #240, #254.

rubidge:
## Motivation and Context
* For Class types, interpreted or compiled, support magic methods, except for `__getattribute__`, `__complex__`, `__enter__`, `__exit__`.

## Approach
* Fix inplace default fallback for Alternative and Class types.
* Support reverse operator magic methods (`__radd__`, etc.) for Alternative and Class types.
* Not supported: `__getattribute__`, `__complex__`, `__enter__`,` __exit__`.
* Avoid Alternative (and Class) types inadvertently converting to float or int.
    * Test case is test_compile_alternative_float_conv (and test_compile_class_float_conv).
    * Want numerics (including numpy) to convert, but Alternative (and Class) to require a cast.
* Fix interpreted default Alternative len().
* Fix interpreted `__setitem__` for Alternative types.
* For ConstDict, check types for + and - operators (interpreted).
* Fix TupleOfListOfInstance operator not implemented behavior.
* Add bool conversions for PointerTo, but not testable yet.
* Compile Dict.get with 2 parameters.  Refs #254.

## How Has This Been Tested?
* Tests were written first.

## Types of changes

## Checklist:

### DIFF
--- a/typed_python/PyClassInstance.cpp
+++ b/typed_python/PyClassInstance.cpp
@@ -149,7 +149,7 @@ int PyClassInstance::classInstanceSetAttributeFromPyObject(Class* cls, instance_
 }
 
 PyObject* PyClassInstance::pyUnaryOperatorConcrete(const char* op, const char* opErr) {
-    std::pair<bool, PyObject*> res = callMemberFunction(op);
+    auto res = callMemberFunction(op);
 
     if (!res.first) {
         return PyInstance::pyUnaryOperatorConcrete(op, opErr);
@@ -159,13 +159,22 @@ PyObject* PyClassInstance::pyUnaryOperatorConcrete(const char* op, const char* o
 }
 
 PyObject* PyClassInstance::pyOperatorConcrete(PyObject* rhs, const char* op, const char* opErr) {
-    std::pair<bool, PyObject*> res = callMemberFunction(op, rhs);
+    auto res = callMemberFunction(op, rhs);
 
-    if (!res.first) {
-        return PyInstance::pyOperatorConcrete(rhs, op, opErr);
+    if (res.first) {
+        return res.second;
     }
 
-    return res.second;
+    if (strlen(op) > 2 && strlen(op) < 16 && op[2] == 'i') { // an inplace operator should fall back to the regular operator
+        char reg_op[16] = "__";
+        strcpy(&reg_op[2], op + 3);
+
+        auto res = callMemberFunction(reg_op, rhs);
+        if (res.first)
+            return res.second;
+    }
+
+    return PyInstance::pyOperatorConcrete(rhs, op, opErr);
 }
 
 PyObject* PyClassInstance::pyOperatorConcreteReverse(PyObject* lhs, const char* op, const char* opErr) {
@@ -174,7 +183,7 @@ PyObject* PyClassInstance::pyOperatorConcreteReverse(PyObject* lhs, const char* 
     buf[0] = '_';
     buf[2] = 'r';
 
-    std::pair<bool, PyObject*> res = callMemberFunction(buf, lhs);
+    auto res = callMemberFunction(buf, lhs);
 
     if (!res.first) {
         return PyInstance::pyOperatorConcreteReverse(lhs, buf, opErr);
@@ -183,7 +192,7 @@ PyObject* PyClassInstance::pyOperatorConcreteReverse(PyObject* lhs, const char* 
     return res.second;
 }
 
-PyObject* PyClassInstance::pyTernaryUnaryOperatorConcrete(PyObject* rhs, PyObject* ternaryArg, const char* op, const char* opErr) {
+PyObject* PyClassInstance::pyTernaryOperatorConcrete(PyObject* rhs, PyObject* ternaryArg, const char* op, const char* opErr) {
     if (ternaryArg == Py_None) {
         //if you pass 'None' as the third argument, python calls your class
         //__pow__ function with two arguments. This is the behavior for
@@ -191,7 +200,7 @@ PyObject* PyClassInstance::pyTernaryUnaryOperatorConcrete(PyObject* rhs, PyObjec
         return pyOperatorConcrete(rhs, op, opErr);
     }
 
-    std::pair<bool, PyObject*> res = callMemberFunction(op, rhs, ternaryArg);
+    auto res = callMemberFunction(op, rhs, ternaryArg);
 
     if (!res.first) {
         return PyInstance::pyTernaryOperatorConcrete(rhs, ternaryArg, op, opErr);
@@ -202,9 +211,9 @@ PyObject* PyClassInstance::pyTernaryUnaryOperatorConcrete(PyObject* rhs, PyObjec
 
 int PyClassInstance::pyInquiryConcrete(const char* op, const char* opErrRep) {
     // op == '__bool__'
-    std::pair<bool, PyObject*> p = callMemberFunction("__bool__", nullptr, nullptr);
+    auto p = callMemberFunction("__bool__");
     if (!p.first) {
-        p = callMemberFunction("__len__", nullptr, nullptr);
+        p = callMemberFunction("__len__");
         // if neither __bool__ nor __len__ is available, return True
         if (!p.first)
             return 1;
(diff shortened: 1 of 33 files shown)


typed_python/PyClassInstance.cpp
```diff
@@ -202,17 +211,28 @@ PyObject* PyClassInstance::pyTernaryUnaryOperatorConcrete(PyObject* rhs, PyObjec
...
             return 1;
     }
     return PyObject_IsTrue(p.second);
 }
 
-std::pair<bool, PyObject*> PyClassInstance::callMemberFunction(const char* name, PyObject* arg0, PyObject* arg1) {
+// try to call user-defined hash method
+// returns -1 if not defined or if it returns an invalid value
+int64_t PyClassInstance::tryCallHashMemberFunction() {
+    auto result = callMemberFunction("__hash__");
```
braxtonmckee, line 227: if __hash__ returns `-1` we need to return `-2`.

typed_python/PyClassInstance.cpp
```diff
@@ -332,6 +347,16 @@ void PyClassInstance::constructFromPythonArgumentsConcrete(Class* classT, uint8_
 }
 
 PyObject* PyClassInstance::tp_getattr_concrete(PyObject* pyAttrName, const char* attrName) {
+    auto p = callMemberFunction("__getattribute__", pyAttrName);
```
braxtonmckee: we need a `mHasGetAttributeMagicMethod` flag on the class when it's defined. Otherwise we have to do an extra lookup in a std::map each time we access a class attribute...

typed_python/PyConstDictInstance.cpp
```diff
@@ -234,6 +234,8 @@ PyObject* PyConstDictInstance::pyOperatorConcrete(PyObject* rhs, const char* op,
                 type()->addDicts(dataPtr(), w_rhs->dataPtr(), data);
             });
         } else {
+            if (!PyDict_Check(rhs))
```
braxtonmckee, line 237: as written, we can't add two ConstDict's of different type, but we can add a ConstDict and a regular dict. We should allow adding anything with the mapping protocol to a ConstDict. However, I wouldn't fix this in this commit - let's file a ticket for it and come back to it.

typed_python/PyInstance.cpp
```diff
@@ -316,12 +316,17 @@ PyObject* PyInstance::pyUnaryOperator(PyObject* lhs, const char* op, const char*
...
         if (extractTypeFrom(lhs->ob_type)) {
-            return specializeForType(lhs, [&](auto& subtype) {
+            ret = specializeForType(lhs, [&](auto& subtype) {
                 return subtype.pyOperatorConcrete(rhs, op, opErrRep);
             });
         }
 
+        if (ret && ret != Py_NotImplemented) {
+            return ret;
+            }
```
braxtonmckee: formatting on the `}`

typed_python/PyInstance.cpp
```diff
@@ -587,13 +597,15 @@ PyTypeObject* PyInstance::typeObj(Type* inType) {
...
+            t->getTypeCategory() == Type::TypeCategory::catConcreteAlternative ||
+            t->getTypeCategory() == Type::TypeCategory::catClass
+            ) {
         PySequenceMethods* res =
             new PySequenceMethods {0,0,0,0,0,0,0,0};
             res->sq_contains = (objobjproc)PyInstance::sq_contains;
             res->sq_length = (lenfunc)PyInstance::mp_and_sq_length;
-            res->sq_item = (ssizeargfunc)PyInstance::sq_item;
-            res->sq_ass_item = (ssizeobjargproc)PyInstance::sq_ass_item;
+            //res->sq_item = (ssizeargfunc)PyInstance::sq_item;
```
braxtonmckee: i'm curious why do we not have these any more? are they superseded by other methods or something?

typed_python/PySetInstance.cpp
```diff
@@ -601,7 +601,7 @@ void PySetInstance::copyConstructFromPythonInstanceConcrete(SetType* setType, in
             setType->destroy(tgt);
         }
 
-        throw;
+        throw PythonExceptionSet();
```
braxtonmckee: why are we recasting this as a PythonExceptionSet? If the internals threw a std::runtime_error this will not work correctly.

typed_python/compiler/tests/alternative_compilation_test.py
```diff
@@ -1075,7 +1217,7 @@ def f_delattr2(x: A):
             self.assertEqual(f_getattr2(v), "4")
             self.assertEqual(f_getattr2(v), c_getattr2(v))
             f_delattr1(v)
-            # exception types are different
+            # TODO: exception types are different.  Should this be made consistent?
```
braxtonmckee: yes. we used to not be able to raise anything but TypeError, so our old tests did the wrong thing. But now we can throw real exceptions, so we should be trying to get this right.

typed_python/compiler/tests/python_object_of_type_compilation_test.py
```diff
@@ -362,6 +365,9 @@ def test_bool_cast_and_conv(self):
             (B4, B4.a(s='')),
             (B4, B4.a(s='a')),
             (B5, B5.a(s='')),
+            # TODO: The line below segfaults in the specialized entrypoint case,
+            # due to attempting set iteration on this type
+            # (PointerTo(Int64), x0.pointerUnsafe(0)),
```
braxtonmckee, line 370: please file a ticket so we don't forget about it.

typed_python/compiler/type_wrappers/class_wrapper.py
```diff
@@ -850,3 +883,156 @@ def generateConstructor(self, context, out, *args):
...
+        return self.generate_method_call(context, magic, (r, l)) \
+            or super().convert_bin_op_reverse(context, r, op, l, inplace)
+
+    def convert_hash(self, context, expr):
+        y = self.generate_method_call(context, "__hash__", (expr,))
+        if y is not None:
+            return y
+        tp = context.getTypePointer(expr.expr_type.typeRepresentation)
+        if tp:
+            return context.pushPod(Int32, runtime_functions.hash_class.call(expr.nonref_expr.cast(VoidPtr), tp))
```
braxtonmckee: the default implementation in 'HeldClassType' just hashes the class pointer. We should be able to do that directly in compiler code rather than hitting a runtime function for it. See `HeldClass::hash`

typed_python/compiler/type_wrappers/runtime_functions.py
```diff
@@ -333,6 +333,12 @@ def unaryPyobjCallTarget(name, retType=Void.pointer()):
     UInt64, Void.pointer(), Void.Pointer(), Int64
 )
 
+class_cmp = externalCallTarget(
```
braxtonmckee, line 336: if __cmp__ is not defined, class comparison throws an exception for ordered comparison, and for EQ/NEQ, it just uses the class pointer. We can do the same thing in compiled code.

braxtonmckee (changes requested):
looks good except for a few minor things. Please address in a separate commit and then i'll take the pair. great work!

typed_python/PyClassInstance.cpp
rubidge: I'll make a test for this case.  It looks like the -1 to -2 transformation is already handled, in the compiled case by tp_hash_to_py_hash, and in the interpreted case by PyInstance::tp_hash.

typed_python/PyInstance.cpp
rubidge: [ ] could be the SequenceMethod sq_item/sq_ass_item or it could be the MappingMethod mp_subscript/mp_ass_subscript.  The SequenceMethod methods require an integer index, while the MappingMethod methods are less restrictive.  Since I don't know if the Alternative or Class is emulating a Sequence or a Map, I went with the less restrictive signature.

braxtonmckee:
ok, cool, then lets delete them outright.

typed_python/PySetInstance.cpp
rubidge: My mistake here.

typed_python/PyConstDictInstance.cpp
rubidge: Ok, this has been done.

typed_python/PyClassInstance.cpp
rubidge: Ok, this has been done.

typed_python/PyInstance.cpp
rubidge: done

typed_python/compiler/tests/alternative_compilation_test.py
rubidge: This has been done.

typed_python/compiler/tests/python_object_of_type_compilation_test.py
rubidge: Created.

typed_python/compiler/type_wrappers/class_wrapper.py
rubidge: This has been done.

typed_python/compiler/type_wrappers/runtime_functions.py
rubidge: This has been done.

rubidge:
Changes have been made.  Superseded by PR #262.

PR #266, opened 2019-10-31
Remove convert_cast.  Refs #265.

rubidge:
## Motivation and Context
Treat these casts like operators.

## Approach
Replace convert_cast with specific convert_*_cast for each builtin type, which can then be implemented by types that support those casts.
convert_bool_cast
convert_int_cast
convert_float_cast
convert_str_cast
convert_bytes_cast

Not meant to change any functionality.

## How Has This Been Tested?
Check that these 5 casts (applied to a variety of types) behave the same for compiled and interpreted code.

## Types of changes

## Checklist:


typed_python/compiler/python_object_representation.py
```diff
@@ -175,6 +180,21 @@ def pythonObjectRepresentation(context, f):
     if f is abs:
         return TypedExpression(context, native_ast.nullExpr, AbsWrapper(), False)
 
+    if f is bool:
+        return TypedExpression(context, native_ast.nullExpr, BoolCastWrapper(), False)
```
braxtonmckee: I don't think we want to have separate 'BytesCastWrapper', as this is the representation for 'bytes' itself as a free object. Before your change here, this would be 'PythonTypeObjectWrapper(f)' with 'f=bytes' (or whichever type). So you probably ought to go to the 'def convert_call(self, context, left, args, kwargs):' method of PythonTypeObjectWrapper and modify that code to dispatch to your convert_cast_X operations for these specific subclasses.

Alternatively, you could make subclasses of PythonTypeObjectWrapper called 'BytesTypeObjectWrapper', etc, and have the dispatch there.

braxtonmckee (changes requested):

typed_python/compiler/python_object_representation.py
rubidge: This has been done.  Makes things a lot simpler.

typed_python/_runtime.cpp
```diff
@@ -1074,4 +1074,60 @@ extern "C" {
         }
     }
 
+    int64_t np_str_to_int64(StringType::layout* s) {
+        PyEnsureGilAcquired getTheGil; // since to_int64 can raise ValueErr
```
braxtonmckee: this should be closer to the raise. most of the time we won't be raising, and this is very slow to have in the fastpath.

typed_python/_runtime.cpp
```diff
@@ -1074,4 +1074,60 @@ extern "C" {
...
+
+        return StringType::to_int64(s);
+    }
+
+    double np_str_to_float64(StringType::layout* s) {
```
braxtonmckee, line 1135: we need to get the gil out of this loop. I think you can probably use `PyOS_string_to_double` which doesn't  hit the gil. see 

https://github.com/python/cpython/blob/279f44678c8b84a183f9eeb85e0b086228154497/Objects/floatobject.c

braxtonmckee (commented):
see my comment above about getting the GIL out of the conversion routines and then this is good to go. the str->float stuff also gets rid of _'s. you can see python has a conversion-function-agnostic routine for doing this (whose logic we can just replicate) here:

https://github.com/python/cpython/blob/4fa7504ee3184cff064e23fe6799e717ed0f9357/Python/pystrtod.c#L385

typed_python/_runtime.cpp
rubidge: Ok, working on this, still have a few things to correct.

rubidge:
Superseded by PR #271.

Issue #267, opened 2019-11-01
Compile "and", "or" operators with non-boolean inputs


rubidge:
Example:
Interpreted:
```
0 or "" or 1.5 = {float} 1.5
```

However, currently "or" is forced to a boolean return value, as shown below:

```
        @Compiled
        def f(x: int, y: str, z: float):
            return x or y or z

f(0,"",1.5) = {bool} True
```

Change compiled behavior to match interpreted behavior.

The return type for the case above should be `OneOf(int, str, float)`.

PR #272, opened 2019-11-06
Support compiled "and" and "or" that return non-boolean values.  Refs #267.

rubidge:
## Motivation and Context
This is handled properly in interpreted code.  E.g. `3 and "x"` returns `"x"`
Compiled code needs to be consistent with interpreted code.

## Approach
* Adjusted the compilation of BoolOp to calculate the return type and to return actual values instead of boolean constants.
*  `T1 and T2` has return type `OneOf(T1, T2`).  Same for `T1 or T2`.
* The conversion T1->OneOf(T1,T2) exposed some issues with isReference requirements in OneOfWrapper.  I worked around these issues in a direct way, but don't trust my approach.

## How Has This Been Tested?
Tests for same types, different types, 2 or 3 parameters, "and" and "or".

## Types of changes

## Checklist:

### DIFF
--- a/typed_python/compiler/expression_conversion_context.py
+++ b/typed_python/compiler/expression_conversion_context.py
@@ -188,9 +188,6 @@ class ExpressionConversionContext(object):
             if isMove:
                 returnTarget.expr.store(expression.nonref_expr)
             else:
-                # TODO: Is this really what I want to do?
-                if not expression.isReference:
-                    expression = self.pushMove(expression)
                 returnTarget.convert_copy_initialize(expression)
 
             self.pushTerminal(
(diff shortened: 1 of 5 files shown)


typed_python/compiler/expression_conversion_context.py
```diff
@@ -187,6 +188,9 @@ def pushReturnValue(self, expression, isMove=False):
             if isMove:
                 returnTarget.expr.store(expression.nonref_expr)
             else:
+                # TODO: Is this really what I want to do?
+                if not expression.isReference:
+                    expression = self.pushMove(expression)
```
braxtonmckee: you shouldn't have to do this. OneOf should never not be a reference because it's not a 'pass_by_value' type.

typed_python/compiler/expression_conversion_context.py
```diff
@@ -777,28 +781,32 @@ def convert_expression_ast(self, ast):
             return pythonObjectRepresentation(self, ast.s)
 
         if ast.matches.BoolOp:
-            def convertBoolOp(depth=0):
+            with self.subcontext():
+                return_type = OneOfWrapper.mergeTypes((self.convert_expression_ast(v).expr_type.typeRepresentation for v in ast.values))
+
+            def convertBoolOp(return_type, depth=0):
```
braxtonmckee: you don't need to pass return_type here - it's captured in the parent closure.

typed_python/compiler/expression_conversion_context.py
```diff
@@ -777,28 +781,32 @@ def convert_expression_ast(self, ast):
...
                                 sc.expr = native_ast.Expression.Branch(
-                                    cond=value.nonref_expr, true=native_ast.trueExpr, false=tail_expr)
+                                    cond=bool_value.nonref_expr, true=value_expr, false=tail_expr)
                             else:
                                 raise Exception(f"Unknown kind of Boolean operator: {ast.op.Name}")
 
                 return sc.result
 
-            return TypedExpression(self, convertBoolOp(), typeWrapper(bool), False)
+            return TypedExpression(self, convertBoolOp(return_type), typeWrapper(return_type), False)
```
braxtonmckee: this is the source of the problem. the 'branch' statements above are selecting between different versions of 'value_expr', which are themselves reference types. But then we create a TypedExpression (whose value is a non-reference) out of the merge of the different pathways. this isn't supposed to happen, because we need OneOf's to be reference to be able to manipulate them correctly.

braxtonmckee (commented):
I pushed up a commit that appears to work. This is very finicky code, so maybe add some more tests. In particular we should check that we are correctly handling the type inference when one of the subexpressions always throws an exception.

typed_python/compiler/expression_conversion_context.py
rubidge: Thanks, will add more tests.

rubidge: There's a problem with the test I committed.  Working on it. [Edit: False alarm -- it's ok.]

braxtonmckee (approved):

PR #330, opened 2020-05-29
Support all functions in 'math' module.  Refs #328, #329, #331.

### LINKED ISSUE #328: Port over remaining members of 'math' module.

### LINKED ISSUE #329: @Entrypoint and int parameter: selection of overload is not consistent.
```
        @Entrypoint
        def f(x):
            return (type(x), int(x))

        r1 = f(UInt8(-1))
        r2 = f(-1)
```
produces
```
(<class 'UInt8'>, 255)
(<class 'UInt8'>, 255)
```

It should produce
```
(<class 'UInt8'>, 255)
(<class 'int'>, -1)
```
instead.

The result depends on the call sequence.  If the 'int' overload appears first:
```
        r1 = f(-1)
        r2 = f(UInt8(-1))
```
the result is ok:
```
(<class 'int'>, -1)
(<class 'UInt8'>, 255)
```

This is not a problem for other integer types besides 'int'.
This is not a problem if we define the overloads with @Compiled instead of @Entrypoint:
```
        @Compiled
        def f(x: UInt8):
            return (type(x), int(x))

        @Compiled
        def f(x: int):
            return (type(x), int(x))
```
In general, should we use the sign or size of an int value to select the overload?

### LINKED ISSUE #331: Float32 overload is selected unexpectedly
Siimilar to #329, but with Float32 and float

```
        @Entrypoint
        def f(x):
            return type(x)(x + 1e-8)

        @Entrypoint
        def g(x):
            return type(x)(x + 1e-8)

        r1 = f(1.0)
        r2 = f(Float32(1.0))
        print(r1, r2)

        r3 = g(Float32(1.0))
        r4 = g(1.0)
        print(r3, r4)
```

output is 

```
1.00000001 1f32
1f32 1f32
```

rubidge:
## Motivation and Context
Previously, only a few functions in the math module were compilable.

## Approach
* Some compiled functions are implemented with llvm intrisic functions.
* Others are implemented with C++.
* A few simple ones are implemented with native_ast operations
* fsum is implemented as a compiled python function.

* factorial has been defined for both integers and floats:
  e.g. factorial(20) is an int
       factorial(Float32(20.0)) is a Float32.
       factorial(20.0) is a float.
* factorial for integers handles up to math.factorial(20).
* factorial for Float32 and float factorial() handles larger numbers
  This allows larger values, up to factorial(170.0)

* Also, fixed some overload selection problems with int and float:
    Did this with adjustments to pyValCouldBeOfTypeConcrete.

## Types of changes

## Checklist:

### DIFF
--- a/typed_python/PyRegisterTypeInstance.hpp
+++ b/typed_python/PyRegisterTypeInstance.hpp
@@ -1,5 +1,5 @@
 /******************************************************************************
-   Copyright 2017-2019 typed_python Authors
+   Copyright 2017-2020 typed_python Authors
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
@@ -582,7 +582,7 @@ public:
     static bool pyValCouldBeOfTypeConcrete(modeled_type* t, PyObject* pyRepresentation, bool isExplicit) {
         if (isFloat(t->getTypeCategory()))  {
             if (PyFloat_Check(pyRepresentation)) {
-                return true;
+                return isExplicit || t->getTypeCategory() == Type::TypeCategory::catFloat64;
             }
 
             if (!isExplicit) {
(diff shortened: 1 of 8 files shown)


braxtonmckee:
Great work!

PR #333, opened 2020-06-04
Fix Set initialization from ListOf or TupleOf.  Refs #332.

### LINKED ISSUE #332: Set doesn't initialized as expected
```
len(Set(int)(ListOf(int)((1,1))))==2
```
returns True...

rubidge:
## Motivation and Context
When a Set type was initialized with a ListOf or TupleOf type, each element was inserted, without checking for duplicates.  So the constructed Set was invalid.

## Approach
Changed ListOf and TupleOf initialization for a Set instance to match initialization from an iterable.

## How Has This Been Tested?
Check len after initializing with duplicate elements.

## Types of changes

## Checklist:


braxtonmckee (approved):
great!

PR #345, opened 2020-06-30
Compile list, set, and dict comprehensions.  Refs #287, 342, 344.

### LINKED ISSUE #287: Allow compilation of list/set/dict comprehensions and generators.

### LINKED ISSUE #342: String iteration doesn't compile.
```
        def f0(s):
            for c in s:
                if c == ' ':
                    return True
            return False
```

`Entrypoint(f0)("a")` produces

`{attributeError}__iter__`

### LINKED ISSUE #344: Dict missing some methods
If `d = Dict(...)` 
* When interpreted, d.copy() generates `AttributeError:'Dict(...)' object has no attribute 'copy'`.
* `d.__setitem__(k,v)` does not compile.  (But `d[k]=v` compiles.)

I'll have to check to see if there are other missing operations.

rubidge:
## Motivation and Context
This was missing functionality.
Please note: nested comprehensions are not supported yet.
Please note: This PR is built on top of PR #341 (i.e. it includes all the changes of PR #341)

## Approach
Comprehensions are compiled by transforming the ast into a form that uses more elementary operations, which are then compiled as usual.  This transform requires knowing the element type of an iterator, so some out-of-band (not via compilation) determination of the element type was implemented in the wrapper.  For example, code was added so that we know if we iterate over:
* a ListOf(T), we get a T.
* a PythonObjectOfType(range), we get an int.
* etc.

The type of a comprehension is a ListOf, Set, or Dict, masqueraded to list, set, or dict, respectively.

Along the way, some fixes were done to Dict operations (Refs #344), and to str operations (Refs #342).
Also, str*int multiplication is now compilable.
These changes made some test cases easier to write.

## How Has This Been Tested?
* multiple iterators are supported
* multiple conditions are supported
* different expression types
* result type is just 'list', 'set', or 'dict'

## Types of changes

## Checklist:

### DIFF
--- a/typed_python/PyDictInstance.cpp
+++ b/typed_python/PyDictInstance.cpp
@@ -200,6 +200,47 @@ PyObject* PyDictInstance::dictClear(PyObject* o) {
     return incref(Py_None);
 }
 
+void copy_elements(PyObject* dst, PyObject* src) {
+    PyDictInstance* dst_w = (PyDictInstance*)dst;
+    iterate(src, [&](PyObject* key) {
+        PyObjectHolder value(PyObject_GetItem(src, key));
+        if (!value) {
+            throw PythonExceptionSet();
+        }
+
+        if (dst_w->mp_ass_subscript_concrete(key, value) == -1) {
+            throw PythonExceptionSet();
+        }
+    });
+}
+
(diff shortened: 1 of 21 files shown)


braxtonmckee:
a full generator model is implemented in 'dev' now.

PR #348, opened 2020-07-10
Gus getitem

### LINKED ISSUE #347: Can't negatively index Tuples/NamedTuples inside Entrypoint
```
@Entrypoint
def f(x):
    return x[-1]

f(Tuple(int)((0,)))
```
raises an IndexError. Same with NamedTuple(a=int)(a=0).

affine-root-system:
Allow tuples and named tuple to be sliced on negative integers (up to, and including, minus the length of the tuple). Previously this failed under entrypoint.

Also allow expressions like '-1' to be interpreted as constants by the compiler. Previously it was treated as a typed expression consisting of a constant '1' and a native ast 'Negate'. Before this change, the compiler would not know the output type of 'tup[-1]' for a tuple tup beyond OneOf(*type(tup).ElementTypes).

Also includes a minor change where constant typed expressions are treated by the compiler without direct reference to their internals (a trivial change since they already carry their associated python constants on the surface).

Fixes #347 

## Types of changes

## Checklist:

### DIFF
--- a/typed_python/compiler/type_wrappers/arithmetic_wrapper.py
+++ b/typed_python/compiler/type_wrappers/arithmetic_wrapper.py
@@ -328,12 +328,20 @@ class IntWrapper(ArithmeticTypeWrapper):
 
     def convert_unary_op(self, context, left, op):
         if op.matches.Not:
+            if left.isConstant:
+                return context.constant(not left.constantValue)
             return context.pushPod(bool, left.nonref_expr.logical_not())
         if op.matches.Invert:
+            if left.isConstant:
+                return context.constant(~left.constantValue)
             return context.pushPod(self, left.nonref_expr.bitwise_not())
         if op.matches.USub:
+            if left.isConstant:
+                return context.constant(-left.constantValue)
             return context.pushPod(self, left.nonref_expr.negate())
         if op.matches.UAdd:
+            if left.isConstant:
+                return context.constant(left.constantValue)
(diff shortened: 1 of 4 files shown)


typed_python/compiler/type_wrappers/arithmetic_wrapper.py
```diff
@@ -328,12 +328,20 @@ def convert_builtin(self, f, context, expr, a1=None):
...
+            if left.isConstant:
+                return context.constant(~left.constantValue)
             return context.pushPod(self, left.nonref_expr.bitwise_not())
         if op.matches.USub:
+            if left.isConstant:
+                return context.constant(-left.constantValue)
             return context.pushPod(self, left.nonref_expr.negate())
         if op.matches.UAdd:
+            if left.isConstant:
+                return context.constant(left.constatValue)
```
braxtonmckee: can we try to get tests that exercise this code? because 'constatValue' won't work...

braxtonmckee (changes requested):

guslonergan:
@braxtonmckee  Fixed your complaint

typed_python/compiler/type_wrappers/arithmetic_wrapper.py
affine-root-system: Added tests to check all the unary ops on constant ints, floats, bools are behaving as expected (and fixed the typo)

PR #353, opened 2020-07-23
Gus getitem

affine-root-system:
Let the compiler understand that unary ops on arithmetic constants are still constants. Also allow the slicing of Tuples on negative integers.

## Motivation and Context

Previously, compiler couldn't see that x[-1] would be a float for x of type Tuple(int, float). That's because -1 was not treated as a constant, so the compiler just thought it was any old int. So the output type was OneOf(int, float). Moreover, it couldn't actually do the slice at runtime because it didn't generate code for slicing by non-constant negative ints. Fixed both of these bugs.

## Types of changes

## Checklist:


braxtonmckee:
some commits taken. leaving some of the mutable tuple stuff behind.

Issue #362, opened 2020-09-23
str slice is sometimes corrupted; str comparison bug


rubidge:
The StringType::getsubstr operation is corrupting multibyte characters (not copying enough bytes).

Interpreted StringType comparisons are erroneous when our StringType doesn't use smallest possible bytes_per_codepoint. Rather than force this constraint, could just add more detailed comparison logic.

(These have been addressed in branch jeff-str, not merged yet.  This issue was entered for recordkeeping.)

rubidge:
The string comparison bug exhibited when comparing containers of strings, e.g. comparing ListOf(str) with of python list of str, and one of the StringType strings was not in lowest bytes_per_codepoint form.

braxtonmckee:
closed by #363

Issue #355, opened 2020-08-03
Chained comparisons aren't compilable


rubidge:
While writing bytes.isalpha as a py function to be compiled, I wanted to be able to use a chained comparison x < y < z where each expression is evaluated at most once.

braxtonmckee:
Closed by #363

Issue #351, opened 2020-07-18
Most `bytes` methods are uncompiled.


braxtonmckee:

braxtonmckee:
Closed by #363

PR #363, opened 2020-09-26
Compile 'bytes' and 'str' operations. Refs #342, #351, #355, #362.

rubidge:
## Motivation and Context
Most 'bytes' and 'str' methods were uncompiled.

## Approach
- Adjust BytesWrapper and StringWrapper layout to be complete.
- Create string iterator.
- Compile string * int operator.
- Comple bytes.isalnum, isalpha, isdigit, islower, isspace, istitle, isupper,
  startswith, endswith, strip, lstrip, rstrip, lower, upper,
  count, find, rfind, index, rindex,
  split, rsplit, join, replace, *, in, decode,
  partition, rpartition, translate, center, ljust, rjust,
  capitalize, expandtabs, splitlines, swapcase, title, zfill.
- Compile bytes.maketrans and str.maketrans as a instance method and as a static method.
- Fix bytes slice operation.
- Fix bytes.split edge cases.
- For chr, ord: propagate constants when compiled.
- Also, fix and test for compiling min/max with spurious keyword parameter.
- Incidentally, support compiling chained comparisons.
- NOTE: partition and rpartition return Tuples that are not masqueraded.
- NOTE: where parameters could be 'bytes' or 'bytes-like' objects, only 'bytes' is supported.
- Compile str encode from bytes, and decode to bytes.
    * This uses PyUnicodeFromEncodedObject and PyUnicode_AsEncodedString, so it covers all supported codecs and error handlers
    * If needed for performance, we could special-case some encodings/decodings.
    * Implemented utf8 encode in c++, but it is slower than PyUnicode_AsEncodedString(), despite the conversions.
- Compile 'str' functions: capitalize, casefold, center, count, expandtabs, ljust, maketrans, partition,
  rfind, rindex, rjust, rpartition, rsplit, splitlines, swapcase, title, translate, zfill.
- Add range parameters for str.startswith, str.endswith.
- Fix str compilation of upper, lower.
- Tested compiling str.find as a py function instead of a c++ function.  Performance was 4 times worse.
- Clean up and fix bugs in str.find, str.split, bytes.split, bytes.rsplit, bytes.splitlines compilation.
- Fix str.getsubstr corruption (not enough bytes copied).
- Adjust `str.__repr__`
- Allow compilation of bytes.startswith and bytes.endswith with 2 or 3 arguments.
- Only remaining str methods are str.format and str.format_map

## How Has This Been Tested?
- Added tests for each operation, including testing error cases that raise exceptions.
- Improved some existing tests that were missing edge cases (e.g. empty bytes)

## Types of changes

## Checklist:

### DIFF
--- a/typed_python/BytesType.cpp
+++ b/typed_python/BytesType.cpp
@@ -1,5 +1,5 @@
 /******************************************************************************
-   Copyright 2017-2019 typed_python Authors
+   Copyright 2017-2020 typed_python Authors
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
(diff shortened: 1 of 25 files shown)


rubidge:
Supersedes PR #354.

Issue #364, opened 2020-09-28
Builtin math functions don't seem to compile correctly on 'object'


braxtonmckee:
For instance, the following test fails:

```
    def test_math_functions_on_object(self):
        @Entrypoint
        def callCeil(t: object):
            return math.ceil(t)

        assert callCeil(1.5) == 2.0
```

with exception `TypeError: Can't compile '<built-in function ceil>' on instance of type 'object'`

rubidge:
Naively, I was going to just compile this as` o.__ceil__()`.  However, float has no `__ceil__` or `__floor__` methods.  (It does have `__trunc__`.)

rubidge:
For most math functions, any type is allowed if it has a `__float__` method. (Exception: `__trunc__`.)

braxtonmckee:
I think the problem is that we don't have the pathway correctly set for
'PythonObjectOfType'.  That's always a tricky case because 'object' _could_
be a float, or it could be something else.  See
'027c1fea364417057b5d27e25f2f8b05dc60b46f' as an example of fixing a bug
where we handled __index__ correctly for int, float, and classes, but not
for an int/float that the compiler knows as an object.

rubidge:
Thanks! most tests are passing now...

PR #365, opened 2020-10-01
Compile math functions on object arguments.  Refs #364.

rubidge:
## Motivation and Context
Addresses issue# 364.

## Approach
- The functions math.ceil, math.floor, math.trunc have magic methods `__ceil__`, `__floor__`, and `__trunc__`.
   - For these functions, the magic method is attempted first.  
   - For ceil and floor, if there is no magic method, attempt to call `__float__` and call the regular ceil or floor on that value.
   - For trunc, if there is no magic method, it raises TypeError.
- For other math functions with float arguments, if `__float__` is available, evaluate the math function on that value.
- If a math function is passed an object instead of a float argument, we try a non-explicit conversion to float.
- Similar handling for math functions with int arguments.
- Also, improved error checking for other math function compilation (e.g. raising OverflowError for inf when appropriate).

## How Has This Been Tested?
Call all math functions with floats and integers as objects, also some classes with `__float__` and other magic methods to simulate other numeric types.

## Types of changes

## Checklist:

### DIFF
--- a/typed_python/_runtime.cpp
+++ b/typed_python/_runtime.cpp
@@ -226,11 +226,23 @@ extern "C" {
     }
 
     double np_cosh_float64(double d) {
-        return std::cosh(d);
+        double ret = std::cosh(d);
+        if (std::isinf(ret)) {
+            PyEnsureGilAcquired getTheGil;
+            PyErr_Format(PyExc_OverflowError, "math range error");
+            throw PythonExceptionSet();
+        }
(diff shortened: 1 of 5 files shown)


rubidge:
Will revise this and resubmit.
Reason: math functions with integer parameters use` __index__` to get the integer value, not `__int__`.  When I was adding test cases to improve test coverage, found this problem.

PR #366, opened 2020-10-02
Compile math functions on object arguments.  Refs #364.

rubidge:
## Motivation and Context
Addresses issue# 364.

## Approach
- The functions math.ceil, math.floor, math.trunc have magic methods `__ceil__`, `__floor__`, and `__trunc__`.
   - For these functions, the magic method is attempted first.  
   - For ceil and floor, if there is no magic method, attempt to call `__float__` and call the regular ceil or floor on that value.
   - For trunc, if there is no magic method, it raises TypeError.
- For other math functions with float arguments, if `__float__` is available, evaluate the math function on that value.
- If a math function is passed an object instead of a float argument, we try a non-explicit conversion to float.
- Math functions with int arguments behave differently.  For gcd, the arguments are considered valid if `__index__` is available.  For ldexp, only type int is acceptable.
- Also, improved error checking for other math function compilation (e.g. raising OverflowError for inf when appropriate).
- Adjust and fix other argument conversions for compiled math functions.

## How Has This Been Tested?
Call all math functions with floats and integers as objects, also some classes with `__float__` and other magic methods to simulate other numeric types.

## Types of changes

## Checklist:

### DIFF
--- a/docs/typed_python.md
+++ b/docs/typed_python.md
@@ -382,11 +382,12 @@ its tag name as `Name`, and the `NamedTuple` that makes up its body as `ElementT
 * Each `Class` has a list of `MemberNames` and `MemberTypes`, along with `BaseClasses`, `MRO`, and `IsFinal`.
 * Each `Tuple` has its items in `ElementTypes`
 * Each `NamedTuple` has its names in `ElementNames` and its types in `ElementTypes`
-* Each `ConstDict` and `Dict` has its key type as `KeyType` and its value type as `ValueType`.
+* Each `ConstDict` and `Dict` has its key type as `KeyType` and its value type as `ValueType`.  The key type is also in `ElementType`.
 * Each `Function` has a list of the individual function overloads as `overloads`. See `typed_python.internals.FunctionOverload`.
 * Each `PointerTo` has the pointed-to type as `ElementType`.
-* Each `Set` has the item type as `KeyType`
+* Each `Set` has the item type as `KeyType` and also as `ElementType`
 * Each `ListOf` and `TupleOf` has its element type as `ElementType`.
+* (For any container, `ElementType` is intended to give the type of an element that you get as you iterate.)
 
 ## The Compiler
 
--- a/typed_python/AlternativeType.hpp
+++ b/typed_python/AlternativeType.hpp
@@ -1,5 +1,5 @@
 /******************************************************************************
-   Copyright 2017-2019 typed_python Authors
+   Copyright 2017-2020 typed_python Authors
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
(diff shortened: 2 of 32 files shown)


braxtonmckee:
Hey Jeff - looks like this is on the right track but I have some comments.

I rebased this on my new changes that improve the way we do value conversion in both the compiler and the interpreter.  I was reading through the code, and noticed a couple of things.

First, you need more checks to ensure that conversions are not returning None. Any check like 'toInt64()' or 'convert_to_type' might return None if the compiler can tell it will definitely throw an exception. if you assume it doesn't then the compiler itself will thro an exception in that case.  

Secondly, make sure that you are getting the right output types. For instance, it looks to me like as written, ldexp on a Float32, int will return a Float32. But in the interpreter, it gets cast to a Float64, since python doesn't natively have a float32. So our compiler needs to mimic that behavior. This might actually simplify things a bit, since for the most part there's no reason to provide float32 pathways because python doesn't have that.  If you wanted to be fancy, later we could attempt to detect something like `Float32(ldexp(a, b))` and use the float32 pathway in that case. But for now, we should just shoot for getting equivalence with the interpreter.

Also note that in the new type conversion model you can write `arg.convert_to_type(float, ConversionLevel.Implicit)` and get a conversion that will convert integers and floats as you'd expect, but not strings, and that's the preferred method for 'implicit' argument conversion (where the interpreter wouldn't explicitly call 'float' on the object). That's what you should use any time you have a function that wouldn't call the __float__ method on a class, for instance. If that's what you want, you can use `x.toFloat64()`, which is equivalent to `float(x)` as it converts as ConversionLevel.New.  This code will take care of converting things like 'object' to the right type.

to the extent that a builtin math-function distinguishes between integers and floats, and you get an object, I think the best path is to do something like this: first check if the arguments are explicitly int or float and use the right pathway in that case. if you can't tell, convert to OneOf(int, float) using ConversionLevel.Implicit, (which will cast integers to int and floats to float, and everything else will fail), return None if that returns None (which propagates the exception). You could go one of two ways:

One way is to wrap the entire convert_call in a '@Wrapper.unwrapOneOfAndValue' decorator, so that you know you're not getting OneOf or Value objects. You can then simply re-call the convert_call method with your OneOf values, and the resulting code will unpack them to concrete types. The other option is to provide a C implementation that accepts OneOf* instances. Probably in that case you should also accept a OneOf* as an argument in which to place the output, rather than trying to return one directly. That has the benefit that it'll put less load on the compiler, but it's more code to maintain.

In either case, make sure to test cases where these functions get object-typed arguments that _wont_ convert to int/float and make sure that works, and also ensure that passing something like 'str' fails gracefully. Also check passing OneOf, and Value types.

As a final note, please make sure that any of your test cases that loop over a large number of inputs to check an invariant provide detailed information when they fail. They should provide full details of the inputs, the functions called, the expectation we have of the output, and the full traceback of any exceptions that were produced. Even if you're just checking that an exception is a given type, if I break the test and just get told ValueError != TypeError in the test report, with no info about which test case failed, it's a pain to figure out what happened.

rubidge:
Braxton:
Thanks for the feedback, will review this carefully. 
-Jeff

PR #369, opened 2020-10-06
Cleanup and comments.

rubidge:
## Motivation and Context
A lot of code could use documentation, especially docstrings.

## Approach
- docstrings added for functions in bytes_wrapper.py.
  - get feedback for this, then do the same for other wrappers.
- removed inheritance from object as redundant.

## Types of changes

## Checklist:


braxtonmckee (approved):
This looks good. I might suggest focusing more on docstrings for user-facing functions. For instance, the methods that we generate in C have no docstrings at all, so something like PointerTo is completely impenetrable to someone who doesn't know what it does already.

Issue #359, opened 2020-08-25
exception corruption in compiled 'with'


rubidge:
Given
```
        class CM():
            def __enter__(self):
                return self

            def __exit__(self, exc_type, exc_val, exc_tb):
                return True

        def f():
            with CM():
                raise NotImplementedError()
```
this code
```
        def repeat_a():
            for i in range(1000):
                try:
                    f()
                except Exception as e:
                    pass
            return 1/0
```
produces ZeroDivisionError, as expected.

But this code
```
        def repeat_b():
            for i in range(1000):
                try:
                    Compiled(f)()
                except Exception as e:
                    pass
            return 1/0
```
does not produce ZeroDivisionError.  Instead, it produces NotImplementedError x 1000, then 'RecursionError: maximum recursion depth exceeded while calling a Python object',

```
Traceback (most recent call last):
  File "/home/jeff/repos/typed_python/typed_python/compiler/tests/conversion_test.py", line 2444, in f
    raise NotImplementedError()
NotImplementedError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/jeff/repos/typed_python/typed_python/compiler/tests/conversion_test.py", line 2444, in f
    raise NotImplementedError()
NotImplementedError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/jeff/repos/typed_python/typed_python/compiler/tests/conversion_test.py", line 2444, in f
    raise NotImplementedError()
NotImplementedError

During handling of the above exception, another exception occurred:

... (997 more) ...

Traceback (most recent call last):
  File "/home/jeff/repos/typed_python/.venv/bin/pytest", line 8, in <module>
    sys.exit(main())
  File "/home/jeff/repos/typed_python/.venv/lib/python3.6/site-packages/_pytest/config/__init__.py", line 125, in main
    config=config
  File "/home/jeff/repos/typed_python/.venv/lib/python3.6/site-packages/pluggy/hooks.py", line 286, in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
  File "/home/jeff/repos/typed_python/.venv/lib/python3.6/site-packages/pluggy/manager.py", line 93, in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
  File "/home/jeff/repos/typed_python/.venv/lib/python3.6/site-packages/pluggy/manager.py", line 87, in <lambda>
[...]

PR #374, opened 2020-10-20
Fixes for 'try' and 'with' compilation. Refs #359

rubidge:
## Motivation and Context
After a fix to how `__exit__` was handled in interpreted Class types, expected that this could fix some problems observed with compiled `with`.
Also, when compiling 'try' statements, many extra calls to clear_exc_info() were generated.

## Approach
Compile 'with' statement with multiple context managers on one line.  Needed to make certain temporary slots more specific.

When compiling 'try', check if an exception has occurred before calling clear_exc_info().  Needed to make the 'exception occurred' indicator non-specific - we can use a slot with the same name everywhere.

Retested cases that used to fail:
- Repeating exceptions in compiled `with` sometimes were chained inappropriately.
- Can now use an uncompiled context manager in a compiled `with` statement.
- While adding these tests, also re-added tests for compiling multiple context manangers in a single `with` statement.  These previously failed, and continued to fail, but were corrected.

## How Has This Been Tested?
Activated tests that used to fail.

## Types of changes

## Checklist:

### DIFF
--- a/typed_python/compiler/expression_conversion_context.py
+++ b/typed_python/compiler/expression_conversion_context.py
@@ -1313,7 +1313,10 @@ class ExpressionConversionContext:
         )
 
         if clear_exc:
-            nativeExpr = nativeExpr >> runtime_functions.clear_exc_info.call()
+            nativeExpr = nativeExpr >> native_ast.Expression.Branch(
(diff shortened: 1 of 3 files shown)


braxtonmckee:
This is merged

PR #375, opened 2020-10-27
Improvements for compiled math functions to work on object arguments.  Refs #364

rubidge:
## Motivation and Context
Addresses issue# 364. 

## Approach
* Math functions now produce results of type float consistently (never `Float32`).
* Better handling of nonfinite values and exceptions (more consistent with interpreter).
* No longer accidentally convert str to float.
* Define conversion level Math that does implicit numeric conversion and `__float__`
  [and `__index__` for 3.8].  This is not a complete conversion level, since it only
  applies to conversions to float.
* Needed to adjust conversion tests since I moved ConversionLevel.New from 5 to 6.
* Compiling ceil, floor: account for python 3.8 changes.
* Rework MathWrapper so most functions are compiled from a table.

## How Has This Been Tested?

* Tests include `OneOf` and `Value` types. 
* Tests include Classes with and without `__float__`, `__ceil__`, `__index__`, and `__int__`.

## Types of changes

## Checklist:

### DIFF
--- a/typed_python/ConversionLevel.hpp
+++ b/typed_python/ConversionLevel.hpp
@@ -12,6 +12,7 @@ enum class ConversionLevel {
     UpcastContainers,
     Implicit,
     ImplicitContainers,
+    Math,
     New,
 };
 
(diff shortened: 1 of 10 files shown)


rubidge:
superseded

PR #395, opened 2022-09-07
Add llvmlite version pinning in requirements.txt

wllgrnt:
## Motivation and Context
Currently installing typed_python using `pip installl -e  .` (standard Python package editable install) will install an importable version of typed_python but one that doesn't build and that causes pytest failures whenever Entrypoint is used (`RuntimeError: LandingPadInst needs to be in a function with a personality.`) - pinning llvmlite to be compatible with the pip lockfile fixes this.

## Approach
N/A, change trivial

## How Has This Been Tested?
`pytest` from repo root has only performance-related flaky test failures.

## Types of changes

## Checklist:


wllgrnt:
NB - new versions of pipenv have deprecated the `pipenv lock --requirements` syntax used in the PR checks, and use `pipenv requirements` instead

PR #398, opened 2022-09-28
Add IR inspection methods

wllgrnt:
## Motivation and Context
Adds the minimum viable code introspection tooling to Runtime - given a compiled Function, this will check the associated converter for functions with matching names and generate the string representations thereof. Will throw a ValueError if nothing is found (most likely because the relevant Function hasn't been run/compiled yet).

## Approach
The ideal approach would add a method to the jit-compiled code (see numba's inspect_types), however this can't be done with the existing Entrypoint decorator as Functions don't implement the __dict__ method, so setattr() doesn't work. A C++ method on the Functions would generate overhead for little gain. One workaround would be to implement a wrapper and masquerade such that the interpreter got access to a wrapped version of the Function with the associated utility methods, with the compiler ignoring this wrapper (credit to @guslonergan).

Further work - some equivalent of https://python-ast-explorer.com/ for the two ASTs, perhaps using ODB.

## How Has This Been Tested?
Ran test script on a basic sum function with and without compilation, and checked the output.

## Types of changes

## Checklist:

### DIFF
--- a/typed_python/compiler/introspect.py
+++ b/typed_python/compiler/introspect.py
@@ -1,0 +1,132 @@
+#   Copyright 2017-2019 typed_python Authors
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+"""introspect.py
+
+Contains helper functions for exploring the objects and syntax trees generatied in the process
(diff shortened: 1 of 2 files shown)


braxtonmckee:
can you maybe add some tests that verify that you can compile something, call this method, and then get a string back out?

typed_python/compiler/runtime.py
```diff
@@ -392,6 +392,64 @@ def resultTypeForCall(self, funcObj, argTypes, kwargTypes):
         _resultTypeCache[key] = mergeTypeWrappers(possibleTypes)
         return _resultTypeCache[key]
 
+    @staticmethod
+    def getNativeIRString(typedFunc: Function) -> str:
```
braxtonmckee: so i guess this gets every instantiation of the function? Like - what happens if you call an entrypointed method with two different argument types? reads like you'll get both versions.

braxtonmckee: maybe see `resultTypeForCall` in the same class, which shows how to get a callTarget, which has the exact name of the compiled function based on the argument types

braxtonmckee (changes requested):
definitely needs a test exercising the functionality

